PR #23, opened 2024-01-03
A handful of fixes

akx:
1. **Check for Existing Contributions**: There are none.
2. **Link Related Issues**: There are none.
3. **Elaborate Your Changes**: Please see below.
4. **Ultralytics Contributor License Agreement (CLA)**: I have contributed to other Ultralytics repositories already.

---

The commits, in order:

1. This removes `pyproject.toml`, as this project is _not_ a Python project that would require installation. (The `pyproject.toml` file also would not work, as it attempts a dynamic version configuration where no dynamic version configuration source is present.) It also removes the duplicated configuration from `pyproject.toml`.
2. This changes the user attribution for the automated commits to `Ultralytics Actions` instead of @glenn-jocher. (`git`'s `user.name` is _not_ a GitHub username; it should be a realname.) This is good Git hygiene, not to mention that since this will [evidently](https://github.com/ultralytics/yolov5/pull/12570) be used in Ultralytics's open-source projects that welcome external contributions, it sounds amiss that @glenn-jocher would be personally auto-formatting PRs.
3. The `action.yml` file in this repository did not strictly adhere to the [GitHub action schema](https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/github-action.json); the [type for `inputs.default` is `string`](https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/github-action.json#L352-L355). (Your editor should highlight this as you edit the file; mine did.)
4. Finally, there was a simple typo; while Tom Preston-Werner is indeed the author of the TOML file format, the extension isn't his first name. 😁


akx:
The fact that `user.name` is not a GitHub username is evident from e.g. the below squash commit:
Screenshot 2024-01-03 at 9 55 35

akx:
As alluded to in https://github.com/ultralytics/yolov5/pull/12572#discussion_r1440297369, if the intent is that the target repository should configure these tools in its own `pyproject.toml`, then any configuration within the action YAML should be elided instead (and the action made to check that a `pyproject.toml` exists before it acts).

glenn-jocher:
> The fact that `user.name` is not a GitHub username is evident from e.g. the below squash commit: Screenshot 2024-01-03 at 9 55 35

@akx ah, I saw this also the other day, but I think this specific commit was when I experimented with removing my email from action.yml. Once I returned it GitHub seems to know it's me. 

BTW, we have a bot account at https://github.com/ultralytics/UltralyticsAssistant, we probably want to use this account. Let me try to apply it and see what happens.

akx:
> ah, I saw this also the other day, but I think this specific commit was when I experimented with removing my email from action.yml. Once I returned it GitHub seems to know it's me.

Yes, GitHub maps commit identities to GitHub identities via email addresses (and you can make a commit look like it's been authored by anyone else [lest you sign all of your own commits](https://docs.github.com/en/authentication/managing-commit-signature-verification)) – but in any case, it should be `"Glenn Jocher"`, not `glenn-jocher`.

> BTW, we have a bot account at @UltralyticsAssistant with email hello@ultralytics.com

That would be much better, yes 😄

pderrenger:
Thanks for your insights @akx! 😀

You're absolutely correct about GitHub using email addresses to map commit identities. It's great to keep our commit histories clean and properly attributed. We'll definitely consider using the `UltralyticsAssistant` bot account for a more unified approach across our repositories. Appreciate the suggestion and the clarification on commit authorship. 🚀

PR #481, opened 2025-07-01
Add GitHub Models as default inference provider

sgoedecke:
GitHub now offers a free LLM inference API that's [native in GitHub Actions](https://docs.github.com/en/github-models/use-github-models/integrating-ai-models-into-your-development-workflow#using-ai-models-with-github-actions), so long as you set the models: read permission. This PR updates the existing OpenAI utility functions to use GitHub Models as the default provider, while still allowing the use of OpenAI if an API key is provided.

If this isn't something you want to do, no worries - I just figured it'd be nice to let people use the cool AI-driven parts of the action without having to go and set up an OpenAI repository secret.

## 🛠️ PR Summary

### 🌟 Summary
This update adds support for using GitHub Models as an alternative to OpenAI for LLM-powered features in Ultralytics Actions, making it easier and more flexible to run AI tasks. 🤖🔄

### 📊 Key Changes
- Added the ability to use GitHub Models for LLM inference when an OpenAI API key is not provided.
- Updated documentation to explain the new GitHub Models option and required permissions.
- Enhanced the code to automatically choose between OpenAI and GitHub Models based on available credentials.
- Added new tests to ensure correct behavior for both OpenAI and GitHub Models, as well as error handling when no credentials are present.

### 🎯 Purpose & Impact
- **More Flexibility:** Users can now use free GitHub Models for AI tasks, reducing reliance on OpenAI API keys.
- **Easier Setup:** No need to obtain an OpenAI API key if you prefer GitHub Models—just use your GitHub token.
- **Improved Reliability:** The system automatically selects the best available LLM provider, ensuring smoother operation.
- **Better Documentation:** Clearer setup instructions help users get started quickly with either provider.

This makes Ultralytics Actions more accessible and user-friendly for a wider range of projects! 🚀


glenn-jocher:
@sgoedecke super interesting, thanks for the PR! Will review this week.

PR #485, opened 2025-08-08
Default GPT-5 to "minimal" reasoning

glenn-jocher:

### DIFF
--- a/actions/__init__.py
+++ b/actions/__init__.py
@@ -22,4 +22,4 @@
 #     ├── test_summarize_pr.py
 #     └── ...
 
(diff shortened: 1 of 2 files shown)


UltralyticsAssistant:
Merged and looking great—thank you, @glenn-jocher! “Details make perfection, and perfection is not a detail.” — Leonardo da Vinci. Your addition of GPT-5 reasoning_effort and the 0.0.89 update give users fine-grained control over depth vs speed/cost with safe defaults, all while staying backward compatible. Appreciate you elevating Ultralytics Actions and enabling smarter, faster pipelines. 🚀

PR #491, opened 2025-08-19
Assign "bug" label to only issues with sufficient information

Y-T-G:
Issues should only have bug labels if the user had manually filled the bug report template with the complete information.

## 🛠️ PR Summary

### 🌟 Summary
Tightens label handling for new GitHub issues by preventing automatic assignment of the "bug" label unless the proper template is used.

### 📊 Key Changes
- Excludes the "bug" label from auto-suggested labels when the item is an issue.
- Maintains case-insensitive matching for other available labels.

### 🎯 Purpose & Impact
- Ensures users follow the required bug report template for proper diagnostics 🧰.
- Improves triage quality by avoiding mis-labeled issues 🧭.
- Reduces noise for maintainers, leading to faster, more accurate responses ⏱️.

### DIFF
--- a/actions/__init__.py
+++ b/actions/__init__.py
@@ -22,4 +22,4 @@
 #     ├── test_summarize_pr.py
 #     └── ...
 
-__version__ = "0.0.93"
+__version__ = "0.0.94"
--- a/actions/first_interaction.py
+++ b/actions/first_interaction.py
@@ -146,6 +146,7 @@ INSTRUCTIONS:
 4. Only use the "Alert" label when you have high confidence that this is an inappropriate {issue_type}.
 5. Respond ONLY with the chosen label names (no descriptions), separated by commas.
 6. If no labels are relevant, respond with 'None'.
+{'7. Only use the "bug" label if the user provides a clear description of the bug, their environment with relevant package versions and a minimum reproducible example.' if issue_type == 'issue' else ''}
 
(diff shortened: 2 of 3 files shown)


glenn-jocher:
@Y-T-G I thought this code was supposed to prevent bugs from being labelled as questions and vice versa. Maybe there's a bug in this code?

```python
    if "bug" in current_labels:
        available_labels.pop("question", None)
    elif "question" in current_labels:
        available_labels.pop("bug", None)
```

Y-T-G:
@glenn-jocher I think thay works fine. But the users can just bypass the template, so it can not have any initial labels

https://github.com/ultralytics/ultralytics/issues/20138

image

Y-T-G:
This one has it too:
image

But the user did follow the bug report template. Not sure why the bug label wasn't added on opening automatically.

https://github.com/ultralytics/ultralytics/issues/21732

glenn-jocher:
Oh I see! You're saying you don't want to auto-label any issues as bugs because we have issue templates that already carry bug labels.

Remember this package is public, it's not meant for Ultralytics consumption but for the the open-source community. It looks like we have about 1500 repos using it so if we make a change it must be broadly applicable to all repos, not just the Ultralytics repo.

Even in the Ultralytics org we have repos without issue templates.

Maybe there's a way to generalize a solution or update the prompt etc.

PR #554, opened 2025-10-19
Improve PR Review prompt

glenn-jocher:
## 🛠️ PR Summary

### 🌟 Summary
Simplifies how automated review comments format code suggestions by posting them exactly as provided, improving reliability and consistency. ✨

### 📊 Key Changes
- Removed automatic indentation adjustment for suggested code in `actions/review_pr.py`.
- Suggestions are now inserted verbatim into GitHub’s suggestion block:
  - Before: inferred indentation from the original line and re-indented the suggestion.
  - Now: no indentation manipulation; uses the raw `suggestion` text.

### 🎯 Purpose & Impact
- Reduces formatting errors where auto-indentation could mangle suggestions or cause them to fail to apply in GitHub. ✅
- Ensures suggestions appear exactly as authored by the review tool or developer, improving predictability. 🎯
- Simplifies the code path (fewer lookups and transforms), lowering maintenance complexity and potential edge-case bugs. 🧹
- Minor behavioral shift: suggestions won’t auto-align to surrounding code, so suggestion authors should include any needed indentation. ✍️

### DIFF
--- a/actions/review_pr.py
+++ b/actions/review_pr.py
@@ -10,26 +10,22 @@ from .utils import GITHUB_API_URL, Action, get_completion, remove_html_comments
 REVIEW_MARKER = "🔍 PR Review"
 EMOJI_MAP = {"CRITICAL": "❗", "HIGH": "⚠️", "MEDIUM": "💡", "LOW": "📝", "SUGGESTION": "💭"}
 SKIP_PATTERNS = [
-    r".*\.lock$",
-    r".*-lock\.(json|yaml|yml)$",  # Lock files
-    r".*\.min\.(js|css)$",
-    r".*\.bundle\.(js|css)$",  # Minified
-    r".*dist/.*",
-    r".*build/.*",
-    r".*vendor/.*",
-    r".*node_modules/.*",  # Generated/vendored
-    r".*\.pb\.py$",
-    r".*_pb2\.py$",
-    r".*_pb2_grpc\.py$",  # Proto generated
-    r"package-lock\.json$",
-    r"yarn\.lock$",
-    r"poetry\.lock$",
-    r"Pipfile\.lock$",  # Package locks
-    r".*\.svg$",
-    r".*\.png$",
-    r".*\.jpg$",
-    r".*\.jpeg$",
-    r".*\.gif$",  # Images
+    r"\.lock$",  # Lock files
+    r"-lock\.(json|yaml|yml)$",
+    r"\.min\.(js|css)$",  # Minified
+    r"\.bundle\.(js|css)$",
+    r"(^|/)dist/",  # Generated/vendored directories
+    r"(^|/)build/",
+    r"(^|/)vendor/",
+    r"(^|/)node_modules/",
+    r"\.pb\.py$",  # Proto generated
+    r"_pb2\.py$",
+    r"_pb2_grpc\.py$",
+    r"^package-lock\.json$",  # Package locks
+    r"^yarn\.lock$",
+    r"^poetry\.lock$",
+    r"^Pipfile\.lock$",
+    r"\.(svg|png|jpe?g|gif)$",  # Images
 ]
 
 
@@ -70,7 +66,7 @@ def generate_pr_review(repository: str, diff_text: str, pr_title: str, pr_descri
     filtered_files = {
         path: lines
         for path, lines in diff_files.items()
-        if not any(re.match(pattern, path) for pattern in SKIP_PATTERNS)
+        if not any(re.search(pattern, path) for pattern in SKIP_PATTERNS)
     }
     skipped_count = len(diff_files) - len(filtered_files)
     diff_files = filtered_files


UltralyticsAssistant (dismissed):
## 🔍 PR Review

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)</sub>

Addressed potential markdown breakage and invalid empty suggestion blocks introduced by the simplified suggestion handling. The suggested change adds guardrails for empty content and triple backticks while preserving the intended behavior of posting suggestions.

💬 Posted 1 inline comment

UltralyticsAssistant (dismissed):
## 🔍 PR Review 4

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)</sub>

1) action.yml: Replace the mutable branch install and restore `packages` to `ultralytics-actions` to ensure reproducibility and prevent system-level mutations that can break the runner. 2) actions/review_pr.py: Update directory skip patterns to match nested paths reliably; current patterns may miss directories not at the repo root.

💬 Posted 2 inline comments

action.yml
```diff
@@ -87,15 +87,17 @@ runs:
...
       # Note tomli required for codespell with pyproject.toml
       env:
         INPUTS_SPELLING: ${{ inputs.spelling }}
       run: |
         # For debug
         # sudo env "PATH=$PATH" uv pip install --system --break-system-packages git+https://github.com/ultralytics/actions@main codespell tomli
-        packages="ultralytics-actions"
+        sudo env "PATH=$PATH" uv pip install --system --break-system-packages git+https://github.com/ultralytics/actions@fix-indent codespell tomli
+        # packages="ultralytics-actions"
+        packages=""
```
UltralyticsAssistant: ❗ **CRITICAL**: Avoid installing from a mutable feature branch and modifying the system interpreter. This introduces supply-chain risk, non-reproducibility, and can break the runner. Also, setting `packages=""` prevents `ultralytics-actions` from being installed later, which may cause runtime failures.

**Suggested change:**
```suggestion
        packages="ultralytics-actions"
```

actions/review_pr.py
```diff
@@ -5,10 +5,32 @@
...
 EMOJI_MAP = {"CRITICAL": "❗", "HIGH": "⚠️", "MEDIUM": "💡", "LOW": "📝", "SUGGESTION": "💭"}
+SKIP_PATTERNS = [
+    r".*\.lock$",
+    r".*-lock\.(json|yaml|yml)$",  # Lock files
+    r".*\.min\.(js|css)$",
+    r".*\.bundle\.(js|css)$",  # Minified
+    r"dist/.*",
+    r"build/.*",
+    r"vendor/.*",
+    r"node_modules/.*",  # Generated/vendored
```
UltralyticsAssistant: ⚠️ **HIGH**: The directory skip patterns (`dist/.*`, `build/.*`, `vendor/.*`, `node_modules/.*`) only match when the path starts with those directories due to use of `re.match`. If those directories appear deeper in the path, they won't be skipped. Consider either anchoring to match anywhere (e.g., prefix with `.*/`) or switching to `re.search` where the patterns are evaluated. Also consider case-insensitive image patterns and precompiling regexes for efficiency if called frequently.

**Suggested change:**
```suggestion
        r".*/dist/.*",
        r".*/build/.*",
        r".*/vendor/.*",
        r".*/node_modules/.*",
```

UltralyticsAssistant (dismissed):
## 🔍 PR Review 5

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)</sub>

1) action.yml now performs an unpinned privileged install from a branch and clears the `packages` variable, which is insecure and breaks reproducibility; revert to the stable `ultralytics-actions` package. 2) Directory skip patterns in `SKIP_PATTERNS` are ineffective with `re.match` for nested paths; adjust patterns or use `re.search` to avoid reviewing vendored/generated code. 3) Ensure `remove_html_comments` is available in `.utils` or add a fallback to prevent import/runtime errors.

💬 Posted 3 inline comments

UltralyticsAssistant (dismissed):
## 🔍 PR Review 6

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)</sub>

1) Pin and comment out the debug-only install (action.yml:98) and reinstate deterministic package selection (action.yml:99); also avoid resetting `packages` to empty (action.yml:100) to prevent install failures. 2) Ensure `remove_html_comments` exists or is safely guarded to avoid ImportError at runtime (actions/review_pr.py:8). 3) Tighten `SKIP_PATTERNS` to avoid false positives and improve performance by anchoring directory boundaries and avoiding redundant `.*` with `re.match` (actions/review_pr.py:12).

💬 Posted 5 inline comments

action.yml
```diff
@@ -95,7 +95,9 @@ runs:
...
```
UltralyticsAssistant: ❗ **CRITICAL**: Resetting `packages` to an empty string risks `uv pip install $packages` failing with 'no requirements' and disables optional installs appended later. Preserve the current value instead of resetting.

**Suggested change:**
```suggestion
        packages="$packages"
```

UltralyticsAssistant (dismissed):
## 🔍 PR Review 7

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)</sub>

- Critical: action.yml installs from a temporary/non-default branch and unconditionally installs optional tools; switch to a stable source and respect inputs. Also ensure `packages` isn't reset to empty, which can break installs.
- Medium: Regex skip patterns combined with `re.match` are redundant/inefficient; consider `re.search` or anchored/compiled patterns.
- Low: New import may fail if `.utils.remove_html_comments` is unavailable; consider a safe fallback. Documentation could better set expectations for skipped files and limited comments.
- Note: Double-check the GitHub Reviews API usage elsewhere in this PR; when posting a single PR review with multiple inline comments, the REST API typically expects `position` (diff-based) for batched comments. Using `line`/`side` may not be supported in that endpoint and could cause 422 errors.

💬 Posted 5 inline comments

action.yml
UltralyticsAssistant: ⚠️ **HIGH**: Resetting `packages` to an empty string may break the downstream logic that aggregates and installs required packages (e.g., `ultralytics-actions`). If subsequent steps expect `packages` to include the core package, this can lead to missing dependencies or a no-op install.

**Suggested change:**
```suggestion
        packages="ultralytics-actions"
```

actions/review_pr.py
```diff
@@ -5,10 +5,32 @@
 import json
 import re
 
-from .utils import GITHUB_API_URL, Action, get_completion
+from .utils import GITHUB_API_URL, Action, get_completion, remove_html_comments
 
 REVIEW_MARKER = "🔍 PR Review"
 EMOJI_MAP = {"CRITICAL": "❗", "HIGH": "⚠️", "MEDIUM": "💡", "LOW": "📝", "SUGGESTION": "💭"}
+SKIP_PATTERNS = [
```
UltralyticsAssistant: 💡 **MEDIUM**: The `SKIP_PATTERNS` are used with `re.match`, but many patterns start with `.*`, which is redundant and inefficient with `re.match` (which matches from the start). Consider either switching to `re.search` and removing leading `.*`, or anchoring the patterns appropriately (e.g., `(^|/)dist/`) and compiling them once to improve performance. Ensure patterns precisely target generated/vendored artifacts to avoid accidental skips.

action.yml
```diff
@@ -95,7 +95,9 @@ runs:
       run: |
         # For debug
         # sudo env "PATH=$PATH" uv pip install --system --break-system-packages git+https://github.com/ultralytics/actions@main codespell tomli
-        packages="ultralytics-actions"
+        sudo env "PATH=$PATH" uv pip install --system --break-system-packages git+https://github.com/ultralytics/actions@fix-indent codespell tomli
```
UltralyticsAssistant, line 98: ❗ **CRITICAL**: This line force-installs from an unpinned Git branch (`fix-indent`) with elevated privileges (`sudo`) and `--break-system-packages`, bypassing the controlled `packages` flow. This creates non-reproducible builds, broadens the attack surface, and alters behavior by installing tools unconditionally. It also risks conflicts with subsequent package handling. Recommend disabling this debug/install line or gating it behind an input and using a pinned release/commit.

**Suggested change:**
```suggestion
        # sudo env "PATH=$PATH" uv pip install --system --break-system-packages git+https://github.com/ultralytics/actions@fix-indent codespell tomli
```

action.yml
```diff
@@ -95,7 +95,9 @@ runs:
...
+        # packages="ultralytics-actions"
```
UltralyticsAssistant, line 99: ❗ **CRITICAL**: Commenting out the `packages="ultralytics-actions"` assignment prevents the core package from being installed via the intended mechanism, likely breaking the action when `INPUTS_SPELLING` is false or when other tools are required. Restore the assignment to ensure the action functions as expected.

**Suggested change:**
```suggestion
        packages="ultralytics-actions"
```

action.yml
UltralyticsAssistant, line 100: ⚠️ **HIGH**: Overriding `packages` to an empty string disables all package installation logic that follows, which can lead to missing dependencies and runtime failures. If the intent was to keep the variable unchanged, avoid resetting it here.

**Suggested change:**
```suggestion
        # packages=""
```

actions/review_pr.py
```diff
@@ -5,10 +5,28 @@
...
```
UltralyticsAssistant, line 12: ⚠️ **HIGH**: The `SKIP_PATTERNS` may miss common cases: (1) Anchored patterns like `^yarn\.lock$` and `^package-lock\.json$` won’t match when these files live in subdirectories (e.g., `web/yarn.lock`). Consider using `(^|/)...$` for such cases. (2) Image extension filtering is case-sensitive, so `IMG.PNG` or `Logo.SVG` won’t be skipped. Consider a case-insensitive match or patterns that account for uppercase. (3) For performance and clarity, compiling these regexes once (e.g., precompiled with `re.compile`) would avoid recompilation on each file. Combine these improvements to reduce false negatives and speed up filtering.

actions/review_pr.py
```diff
@@ -5,10 +5,28 @@
 import json
 import re
 
-from .utils import GITHUB_API_URL, Action, get_completion
+from .utils import GITHUB_API_URL, Action, get_completion, remove_html_comments
```
UltralyticsAssistant, line 8: 💡 **MEDIUM**: New import `remove_html_comments` assumes the symbol exists in `.utils`. If it’s missing or renamed, the action will fail at import time. Ensure `.utils` exports this function (and is covered by tests), or guard usage with a safe fallback.

actions/review_pr.py
```diff
@@ -44,8 +62,20 @@ def generate_pr_review(repository: str, diff_text: str, pr_title: str, pr_descri
     if not diff_files:
         return {"comments": [], "summary": "No files with changes detected in diff"}
 
+    # Filter out generated/vendored files
```
UltralyticsAssistant, line 65: 📝 **LOW**: The `limit = round(128000 * 3.3 * 0.5)` uses multiple magic constants without context. Consider centralizing these as named constants (with brief comments) to clarify intent and avoid accidental regressions if model context sizes or token-to-byte heuristics change.

actions/review_pr.py
```diff
@@ -54,27 +84,41 @@ def generate_pr_review(repository: str, diff_text: str, pr_title: str, pr_descri
...
         "Focus on: Code quality, style, best practices, bugs, edge cases, error handling, performance, security, documentation, test coverage\n\n"
         "FORMATTING: Use backticks for code, file names, branch names, function names, variable names, packages\n\n"
         "CRITICAL RULES:\n"
-        f"1. Generate inline comments with recommended changes for clear bugs/security/syntax issues (up to 10)\n"
-        "2. Each comment MUST reference a UNIQUE line number(s)\n"
-        "3. If a section has multiple issues, combine ALL issues into ONE comment for that section\n"
-        "4. Never create separate comments for the same line number(s)\n"
-        f"5. {priority_guidance}\n\n"
+        "1. Quality over quantity: Zero comments is fine for clean code - only flag truly important issues\n"
+        f"2. {comment_guidance} - these are MAXIMUMS, not targets\n"
```
Borda, line 101: This could be counterproductive; usually, small and focused changes may need more explanation than larger generated code blocks

actions/review_pr.py
```diff
@@ -54,27 +84,41 @@ def generate_pr_review(repository: str, diff_text: str, pr_title: str, pr_descri
...
+        "3. CRITICAL: Do not post separate comments on adjacent/nearby lines (within 10 lines). Combine all related issues into ONE comment\n"
+        "4. When combining issues from multiple lines, use 'start_line' (first line) and 'line' (last line) to highlight the entire range\n"
+        "5. Each comment must reference separate areas - no overlapping line ranges\n"
+        "6. Prioritize: CRITICAL bugs/security > HIGH impact issues > code quality\n"
+        "7. Keep comments concise, friendly, and easy to understand - avoid jargon when possible\n"
+        "8. DO not comment on routine changes: adding imports, adding dependencies, updating version numbers, standard refactoring\n"
```
Borda, line 107: I believe adding a new dependency always has a reason, so maybe not in code via a comment about it in the PR for sure...

actions/review_pr.py
```diff
@@ -54,27 +84,41 @@ def generate_pr_review(repository: str, diff_text: str, pr_title: str, pr_descri
...
+        "9. Trust the developer - only flag issues with clear evidence of problems, not hypothetical concerns\n\n"
```
Borda, line 108:
```
10. If the code and the used variable names are self-explanatory, no other comment is required.
11. If the edited code block already has a comment, revisit it if it still reflects updated code.

PR #557, opened 2025-10-19
Handle missing OPENAI_API_KEY

glenn-jocher:
## 🛠️ PR Summary

### 🌟 Summary
Adds safe-guards to skip OpenAI-powered actions when no API key is available and introduces a helper to detect fork-based PRs. ⚙️🛡️

### 📊 Key Changes
- Introduced `should_skip_openai()` utility to gracefully skip LLM operations when `OPENAI_API_KEY` is missing.
- Applied skip logic in:
  - `first_interaction.py` (auto-labeling and first-response flow)
  - `summarize_pr.py` (PR summary updates)
- Added `is_fork_pr()` helper to detect PRs coming from forks for future conditional handling.

### 🎯 Purpose & Impact
- Prevents action failures in environments without OpenAI credentials, improving reliability and CI stability. ✅
- Avoids noisy errors by cleanly bypassing LLM-dependent steps and logging a clear warning. 🧹
- Lays groundwork to treat forked PRs differently (e.g., security or permission-aware flows). 🔒
- No behavior change for users with valid OpenAI setup; seamless operation continues. 🚀

### DIFF
--- a/actions/utils/github_utils.py
+++ b/actions/utils/github_utils.py
@@ -192,7 +192,7 @@ class Action:
         if not self.pr:
             return False
         head_repo = self.pr.get("head", {}).get("repo", {}).get("full_name")
-        return head_repo and head_repo != self.repository
+        return bool(head_repo) and head_repo != self.repository
 
     def should_skip_openai(self) -> bool:
         """Check if OpenAI operations should be skipped."""


actions/first_interaction.py
```diff
@@ -176,6 +176,9 @@ def get_first_interaction_response(event, issue_type: str, title: str, body: str
 def main(*args, **kwargs):
     """Executes auto-labeling and custom response generation for new GitHub issues, PRs, and discussions."""
     event = Action(*args, **kwargs)
+    if event.should_skip_openai():
+        return
+
```
UltralyticsAssistant, line 181: ⚠️ **HIGH**: Returning early when `OPENAI_API_KEY` is missing skips the entire action, likely bypassing non-LLM logic (e.g., auto-labeling or other processing). To avoid a regression in triage behavior, gate only LLM-dependent calls or provide a non-LLM fallback rather than exiting the function.

actions/utils/github_utils.py
```diff
@@ -187,6 +187,22 @@ def is_org_member(self, username: str) -> bool:
         """Checks if a user is a member of the organization."""
         return self.get(f"{GITHUB_API_URL}/orgs/{self.repository.split('/')[0]}/members/{username}").status_code == 204
 
+    def is_fork_pr(self) -> bool:
+        """Checks if PR is from a fork (different repo than base)."""
+        if not self.pr:
+            return False
+        head_repo = self.pr.get("head", {}).get("repo", {}).get("full_name")
+        return head_repo and head_repo != self.repository
```
UltralyticsAssistant: ⚠️ **HIGH**: `is_fork_pr` is annotated to return `bool`, but `return head_repo and head_repo != self.repository` can return `None` or a non-boolean (string) due to Python's `and` semantics. Ensure a strict boolean is returned to prevent subtle bugs in callers.

**Suggested change:**
```suggestion
        return bool(head_repo) and head_repo != self.repository
```

UltralyticsAssistant (changes requested):
## 🔍 PR Review

<sub>Made with ❤️ by [Ultralytics Actions](https://github.com/ultralytics/actions)</sub>

Good improvement to safely skip LLM operations when the API key is missing. However, avoid exiting the entire workflow so non-LLM behavior still runs, and fix the boolean return in `is_fork_pr` to match its type annotation. Consider adding tests for the missing-key path and fork detection behavior.

💬 Posted 2 inline comments

PR #576, opened 2025-10-21
Prevent recursive commits

glenn-jocher:
## 🛠️ PR Summary

### 🌟 Summary
Prevents bot-triggered commit loops and refines automated PR review keyword detection for more reliable CI behavior. 🤖✅

### 📊 Key Changes
- Added a safeguard in `action.yml` to skip committing if the last commit author matches the bot user, preventing recursive commit loops.
- Updated keyword matching in `actions/review_pr.py` by replacing "need to"/"needs to" with "raise"/"needs" to improve detection of review severity.

### 🎯 Purpose & Impact
- Prevents infinite commit loops from the GitHub Action, reducing noisy runs and protecting repo history. 🔁🛑
- Improves the accuracy of automated PR review summaries and status decisions, reducing false positives/negatives. 🧠✨
- Leads to more stable CI workflows and clearer, more actionable automated feedback for contributors and maintainers. 🚀


action.yml
```diff
@@ -232,6 +232,12 @@ runs:
         GITHUB_USERNAME: ${{ inputs.github_username }}
         GITHUB_EMAIL: ${{ inputs.github_email }}
       run: |
+        # Check if last commit was by the bot to prevent recursive loops
```
UltralyticsAssistant, line 235: 💡 **MEDIUM**: Thanks for adding the recursion guard. One thing to watch out for: `git log -1` returns exit code 128 when the checkout has no commits (fresh repo, local testing, etc.), which would make this whole step fail before it can exit gracefully. Please add a small guard (for example, skip the check when `git rev-parse HEAD` fails) so the workflow still succeeds on empty repos.

PR #586, opened 2025-10-21
Update action.yml

glenn-jocher:
## 🛠️ PR Summary

### 🌟 Summary
Default install source for the action switches from the PyPI package to the GitHub main branch to ensure the latest code is used. 🚀

### 📊 Key Changes
- When not installing from a specific branch, the action now installs via `git+https://github.com/ultralytics/actions@main` instead of the PyPI package `ultralytics-actions`.
- Existing behavior remains for branch builds: if running from a branch (`GITHUB_REF`), it installs that branch directly from GitHub.

### 🎯 Purpose & Impact
- Ensures users always get the most up-to-date fixes and features from the main branch. 🔧
- Reduces delays or mismatches caused by out-of-date PyPI releases. ⏱️
- Potentially less pinned/stable than PyPI, which could affect reproducibility; users may see faster changes and need to monitor updates. ⚠️

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -101,7 +101,7 @@ runs:
           packages="git+https://github.com/ultralytics/actions@${GITHUB_REF#refs/heads/}"
         else
           # packages="ultralytics-actions"
-          "git+https://github.com/ultralytics/actions@main"
+          packages="git+https://github.com/ultralytics/actions@main"
         fi
 
         if [ "$INPUTS_SPELLING" = "true" ]; then


action.yml
```diff
@@ -100,7 +100,8 @@ runs:
           echo "Installing from git branch: $GITHUB_REF"
           packages="git+https://github.com/ultralytics/actions@${GITHUB_REF#refs/heads/}"
         else
-          packages="ultralytics-actions"
+          # packages="ultralytics-actions"
+          "git+https://github.com/ultralytics/actions@main"
```
UltralyticsAssistant: ❗ **CRITICAL**: `packages` no longer receives a value in the default branch path. Later when we call `pip install "$packages"`, the variable will be empty and the install step will fail. Please restore the assignment.

**Suggested change:**
```suggestion
          packages="git+https://github.com/ultralytics/actions@main"
```

PR #607, opened 2025-10-23
Workflow dispatch from PR forks

glenn-jocher:
## 🛠️ PR Summary

### 🌟 Summary
Bumps Actions to 0.2.0 and overhauls fork PR handling by creating a real temp branch via git operations, adds safer HTTP handling, and improves CI comment messaging. 🚀

### 📊 Key Changes
- Version update: `__version__` from 0.1.9 to 0.2.0.
- Fork PR handling: replace API ref creation with secure git workflow:
  - Clone base repo with `GITHUB_TOKEN`, add fork remote, fetch fork branch into a temp branch, and push to base repo.
  - Sanitizes errors to prevent token leakage.
- CI comment text refined for clarity and formatting.
- GitHub API status handling updated:
  - Accepts 204 for GET and POST to avoid false failures on no-content responses.
- Tests updated to reflect new git flow and environment requirements.

### 🎯 Purpose & Impact
- More reliable CI for forked PRs by ensuring the PR code exists as a branch in the base repo ✅.
- Improved security: avoids exposing tokens in error logs and relies on scoped `GITHUB_TOKEN` 🔒.
- Better compatibility with GitHub endpoints that return 204, reducing noisy errors 🧹.
- Slightly longer setup time due to cloning/pushing, but significantly more robust for cross-repo contributions ⏱️➡️🛡️.
- Clearer PR action comments help contributors understand what was triggered and when ✍️.

### DIFF
--- a/actions/dispatch_actions.py
+++ b/actions/dispatch_actions.py
@@ -36,7 +36,7 @@ def get_pr_branch(event) -> tuple[str, str | None]:
             raise ValueError("GITHUB_TOKEN environment variable is not set")
 
         with tempfile.TemporaryDirectory() as tmp_dir:
-            repo_dir = f"{tmp_dir}/repo"
+            repo_dir = os.path.join(tmp_dir, "repo")
             base_url = f"https://x-access-token:{token}@github.com/{base_repo}.git"
             fork_url = f"https://github.com/{fork_repo}.git"
 
@@ -57,13 +57,11 @@ def get_pr_branch(event) -> tuple[str, str | None]:
 
                 # Push temp branch to base repo
                 subprocess.run(["git", "push", "origin", temp_branch], cwd=repo_dir, check=True, capture_output=True)
-            except subprocess.CalledProcessError as exc:
+            except subprocess.CalledProcessError as e:
                 # Sanitize error output to prevent token leakage
-                stderr = exc.stderr.decode() if exc.stderr else "No stderr output"
+                stderr = e.stderr.decode() if e.stderr else "No stderr output"
                 stderr = stderr.replace(token, "***TOKEN***")
-                raise RuntimeError(
-                    f"Failed to create temp branch for fork PR (exit code {exc.returncode}): {stderr}"
-                ) from exc
+                raise RuntimeError(f"Failed to create tmp branch from fork (exit code {e.returncode}): {stderr}") from e
 
         return temp_branch, temp_branch
 


actions/dispatch_actions.py
```diff
@@ -22,12 +26,34 @@ def get_pr_branch(event) -> tuple[str, str | None]:
...
+        token = os.environ.get("GITHUB_TOKEN")
+        if not token:
+            raise ValueError("GITHUB_TOKEN environment variable is not set")
+
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            base_url = f"https://x-access-token:{token}@github.com/{base_repo}.git"
+            fork_url = f"https://github.com/{fork_repo}.git"
+
+            # Clone base repo (minimal)
+            subprocess.run(["git", "clone", "--depth", "1", base_url, tmp_dir], check=True, capture_output=True)
```
UltralyticsAssistant: ❗ **CRITICAL**: `tmp_dir` from `TemporaryDirectory()` already exists, so `git clone … tmp_dir` will fail with “destination path … already exists and is not an empty directory.” Clone into a new child directory (and update the later `cwd` references) so the workflow can handle fork PRs successfully.

UltralyticsAssistant (commented):
## 🔍 PR Review

Please fix the temp directory cloning path so that fork PR handling works; git currently fails because the target directory already exists.

💬 Posted 1 inline comment

actions/dispatch_actions.py
```diff
@@ -22,12 +26,35 @@ def get_pr_branch(event) -> tuple[str, str | None]:
...
+            subprocess.run(["git", "remote", "add", "fork", fork_url], cwd=repo_dir, check=True, capture_output=True)
+            subprocess.run(
+                ["git", "fetch", "fork", f"{fork_branch}:{temp_branch}"],
+                cwd=repo_dir,
+                check=True,
+                capture_output=True,
+            )
+
+            # Push temp branch to base repo
+            subprocess.run(["git", "push", "origin", temp_branch], cwd=repo_dir, check=True, capture_output=True)
```
UltralyticsAssistant: ❗ **CRITICAL**: Embedding the `GITHUB_TOKEN` directly in the git command means that if any of these `subprocess.run(..., check=True)` calls fail, the resulting `CalledProcessError` will include the full command (and token) in logs. That’s a serious secret-leak risk. Please refactor to avoid putting the token in the command string—e.g. inject credentials via `env`, `GIT_ASKPASS`, or catch failures and rethrow with a sanitized message before logging.

UltralyticsAssistant (commented):
## 🔍 PR Review 2

Please address the critical secret-leak risk in `get_pr_branch` by avoiding passing the raw `GITHUB_TOKEN` in the git command; otherwise the token can surface in logs when a subprocess fails.

💬 Posted 1 inline comment

actions/dispatch_actions.py
```diff
@@ -22,12 +26,38 @@ def get_pr_branch(event) -> tuple[str, str | None]:
...
-        )
+        fork_repo = head["repo"]["full_name"]
+        fork_branch = head["ref"]
+        base_repo = event.repository
+        token = os.environ.get("GITHUB_TOKEN")
+        if not token:
+            raise ValueError("GITHUB_TOKEN environment variable is not set")
+
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            repo_dir = f"{tmp_dir}/repo"
```
UltralyticsAssistant: 📝 **LOW**: Using manual string concatenation to build the repo path can introduce mixed separators on Windows runners. Please rely on `os.path.join` here.

**Suggested change:**
```suggestion
            repo_dir = os.path.join(tmp_dir, "repo")
```

UltralyticsAssistant (commented):
## 🔍 PR Review 3

Please keep the subprocess failure details when bubbling up the error so we can debug git issues, and use `os.path.join` for the temporary repo path to stay cross-platform friendly.

💬 Posted 2 inline comments

actions/dispatch_actions.py
```diff
@@ -22,12 +26,43 @@ def get_pr_branch(event) -> tuple[str, str | None]:
...
+        fork_branch = head["ref"]
+        base_repo = event.repository
+        token = os.environ.get("GITHUB_TOKEN")
+        if not token:
+            raise ValueError("GITHUB_TOKEN environment variable is not set")
+
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            repo_dir = os.path.join(tmp_dir, "repo")
+            base_url = f"https://x-access-token:{token}@github.com/{base_repo}.git"
+            fork_url = f"https://github.com/{fork_repo}.git"
```
UltralyticsAssistant, line 41: ⚠️ **HIGH**: Thanks for adding fork support! One follow-up: `git fetch` against the fork will fail for private or same-org forks because this URL lacks authentication, which is exactly the scenario we need to support. Could we authenticate the fork remote with the same token we use for the base clone?

**Suggested change:**
```suggestion
            fork_url = f"https://x-access-token:{token}@github.com/{fork_repo}.git"
```

UltralyticsAssistant (commented):
## 🔍 PR Review 4

Looks solid overall. Just ensure the fork remote is cloned with the token so private/internal fork PRs can be fetched successfully.

💬 Posted 1 inline comment

PR #609, opened 2025-10-24
Allow 2 successive commits in PRs

glenn-jocher:
## 🛠️ PR Summary

### 🌟 Summary
Prevents recursive auto-format commit loops by checking the last two commits and adjusts checkout depth to support this logic. ✅

### 📊 Key Changes
- ⬆️ Increase `actions/checkout` `fetch-depth` from `1` to `2`.
- 🔁 Update auto-format guard:
  - Previously: skipped if the last commit was an auto-format.
  - Now: skips if the last two commits are auto-format, using `git log -2` and a match count.
- 🧹 Improved logging messages to reflect the new check (“last two commits were auto-format”).

### 🎯 Purpose & Impact
- 🛡️ Prevents recursive auto-format commits that can trigger endless CI runs.
- 🔧 More robust detection reduces noise in commit history and avoids wasted CI minutes.
- 🚀 Minimal overhead: slightly deeper fetch with negligible performance impact.
- 🤝 No action required from users; the Action becomes safer and more reliable out-of-the-box.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -124,7 +124,7 @@ runs:
         repository: ${{ github.event.pull_request.head.repo.full_name || github.repository }}
         token: ${{ inputs.token }}
         ref: ${{ github.head_ref || github.ref }}
-        fetch-depth: 1
+        fetch-depth: 2
 
     # File header ------------------------------------------------------------------------------------------------------
     - name: File Header


UltralyticsAssistant (commented):
## 🔍 PR Review

Tighten the commit check by using a fixed-string `grep` and quoting the count before the numeric comparison to keep the action robust.

💬 Posted 2 inline comments

PR #630, opened 2025-10-31
New Google Python Docstring formatter

glenn-jocher:
## 🛠️ PR Summary

### 🌟 Summary
Introduces an Ultralytics-built Python docstring formatter and integrates it into the GitHub Action as an optional step, replacing docformatter for faster, consistent, Google-style docstrings. 🚀

### 📊 Key Changes
- Added a new docstring formatter: `actions/format_python_docstrings.py` with a CLI entry point `ultralytics-actions-format-python-docstrings`.
- Replaced `docformatter` usage with the new formatter across the Action and Markdown code block tooling.
- New Action input: `python_docstrings` (default: false) to toggle docstring formatting.
- Updated workflows and README:
  - `.github/workflows/format.yml` uses `python: true` (Ruff) and `python_docstrings: true/false` to control formatting.
  - README now documents optional Python docstring formatting and removes references to docformatter.
- Removed `docformatter` dependency and config from `pyproject.toml`; added CLI entry point for the new formatter.
- Minor housekeeping: update module index, tweak comments, and ignore `test_docstring_formatter.py`.

### 🎯 Purpose & Impact
- Improves consistency and readability of docstrings with a Google-style formatter tailored by Ultralytics. ✨
- Reduces external dependencies and CI complexity for faster, more reliable formatting runs. ⚡
- Gives users of the Action fine-grained control: enable/disable docstring formatting independently of Ruff. 🔧
- More robust formatting: preserves code fences, tables, lists, and indentation nuances; supports inline docstrings intelligently.
- Potential behavior change: formatting may differ from docformatter; users relying on docformatter-specific behavior should review results when enabling `python_docstrings`. ✅

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -26,7 +26,8 @@ jobs:
         with:
           token: ${{ secrets._GITHUB_TOKEN || github.token }} # Auto-generated fallback
           labels: true # Auto-label issues/PRs using AI
-          python: true # Format Python with Ruff and Ultralytics Python docstring formatter
+          python: true # Format Python with Ruff
+          python_docstrings: true # Format Python docstrings
           prettier: true # Format YAML, JSON, Markdown, CSS
           spelling: true # Check spelling with codespell
           links: false # Check broken links with Lychee
--- a/README.md
+++ b/README.md
@@ -30,8 +30,8 @@ AI-powered formatting, labeling, and PR summaries for Python, Swift, and Markdow
 ### 📄 Features
 
 - **Python Code:** Formatted using [Ruff](https://github.com/astral-sh/ruff), an extremely fast Python linter and formatter
+- **Python Docstrings:** Google-style formatting enforced with Ultralytics Python docstring formatter (optional)
 - **Markdown Files:** Styled with [Prettier](https://github.com/prettier/prettier) to ensure consistent documentation appearance
-- **Docstrings:** Google-style formatting enforced with Ultralytics Python docstring formatter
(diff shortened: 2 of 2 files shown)


actions/format_python_docstrings.py
```diff
@@ -0,0 +1,468 @@
...
+                        continuation.append("")
+                        j += 1
+                    else:
+                        # End of this param
+                        break
+                elif is_param_line(next_line):
+                    # Next param
+                    break
+                else:
+                    # Continuation line
```
UltralyticsAssistant: ⚠️ **HIGH**: `continuation.append(next_line.strip())` collapses any structured continuation lines (bullets, numbered lists, code blocks) into a single paragraph when we later join them with spaces. For example, an argument description containing `* first` and `* second` will be rewritten as `* first * second`, breaking the formatting docformatter previously preserved. We need to keep the original line breaks/indentation for non-paragraph continuation lines so structured content survives the rewrite.

actions/format_python_docstrings.py
```diff
@@ -0,0 +1,468 @@
...
+        if check:
+            return original == formatted
+
+        if original != formatted:
+            path.write_text(formatted, encoding="utf-8")
+            print(f"Formatted: {path}")
+            return False
+
+        return True
+    except Exception as e:
```
UltralyticsAssistant: 💡 **MEDIUM**: Catching every exception and returning `True` means the formatter silently succeeds even when it fails to read or write a file. In `--check` mode this will still exit 0, masking real errors that docformatter used to surface. Please re-raise (or return `False`) on failure so the action can fail fast instead of hiding the problem.

PR #590, opened 2025-10-22
Requests session pooling

glenn-jocher:
## 🛠️ PR Summary

<sub>Made with ❤️ by [Ultralytics Actions](https://www.ultralytics.com/actions)</sub>

### 🌟 Summary
🔄 This PR updates the GitHub API utility to use a persistent `requests.Session()` for HTTP calls, improving connection reuse and aligning tests with the new behavior.

### 📊 Key Changes
- Added a reusable `requests.Session()` object in `actions/utils/github_utils.py` 🧩
- Changed the internal `_request()` method to send all HTTP requests through `self.session` instead of direct `requests` module calls ⚡
- Updated unit tests in `tests/test_github_utils.py` to mock `requests.Session.get` rather than `requests.get` ✅

### 🎯 Purpose & Impact
- Improves efficiency by reusing HTTP connections instead of creating a new one for each request 🚀
- Lays groundwork for more consistent request handling, including easier future support for retries, shared settings, or connection tuning 🛠️
- Keeps test coverage accurate by matching the new session-based implementation 🧪
- For users, this is a low-risk internal improvement with potential performance and reliability benefits, especially for workflows making multiple GitHub API calls 🔒

PR #654, opened 2025-11-16
Reduce PR Review API calls with GraphQL

glenn-jocher:
## 🛠️ PR Summary

### 🌟 Summary
Optimize GitHub PR workflows by batching multiple REST calls into fewer GraphQL requests for PR summaries, reviews, and issue labeling 🚀

### 📊 Key Changes
- 🔁 Reuse existing `pull_request` payload data instead of performing an extra REST call when handling new PR events.
- 🧵 Introduce `get_pr_details_and_labels` to fetch PR body and repository labels in a single GraphQL query for PR-open handling.
- 🧹 Refactor `dismiss_previous_reviews` to use `get_pr_reviews_and_comments` (GraphQL) and `batch_delete_review_comments` to fetch and delete bot reviews/comments more efficiently.
- 🏷️ Add `GRAPHQL_LABEL_AND_COMMENT_ISSUE` and related helpers (`get_multiple_issue_node_ids`, `get_label_ids` usage) to batch-label and comment on closing issues via GraphQL instead of multiple REST calls.
- 📝 Enhance `update_pr_description` to accept a cached PR body, reducing redundant fetches while still safely handling missing descriptions.
- 🔗 Export new GraphQL constants from `actions.utils` and wire them into `summarize_pr.py` and other call sites.

### 🎯 Purpose & Impact
- ⚡ Reduces the number of GitHub REST API calls, lowering latency and the risk of hitting rate limits for busy repositories.
- 🧠 Centralizes PR, review, comment, and label handling through GraphQL, simplifying maintenance and future extensions.
- 🔒 Adds defense-in-depth checks on review/comment authors, improving safety when dismissing or deleting bot-generated content.
- ✅ Keeps existing behavior (summaries, labels, issue auto-labeling) while making the implementation more efficient and scalable for Ultralytics workflows.

### DIFF
--- a/actions/first_interaction.py
+++ b/actions/first_interaction.py
@@ -36,8 +36,7 @@ def get_event_content(event) -> tuple[int, str, str, str, str, str, str]:
         item = data["issue"]
         issue_type = "issue"
     elif name in ["pull_request", "pull_request_target"]:
-        pr_number = data["pull_request"]["number"]
-        item = event.get_repo_data(f"pulls/{pr_number}")
+        item = data["pull_request"]
         issue_type = "pull request"
     elif name == "discussion":
         item = data["discussion"]
@@ -178,8 +177,6 @@ def main(*args, **kwargs):
         return
 
     number, node_id, title, body, username, issue_type, action = get_event_content(event)
-    available_labels = event.get_repo_data("labels")
-    label_descriptions = {label["name"]: label.get("description") or "" for label in available_labels}
 
     # Use unified PR open response for new PRs (summary + labels + first comment in 1 API call)
     if issue_type == "pull request" and action == "opened":
@@ -187,12 +184,18 @@ def main(*args, **kwargs):
             return
 
         print(f"Processing PR open by @{username} with unified API call...")
+        # Fetch PR details and labels in single GraphQL query
+        repo_data = event.get_pr_details_and_labels(number)
+        pr_data = repo_data.get("pullRequest", {})
+        available_labels = repo_data.get("labels", {}).get("nodes", [])
+        label_descriptions = {label["name"]: label.get("description") or "" for label in available_labels}
+
         diff = event.get_pr_diff()
         response = get_pr_open_response(event.repository, diff, title, username, label_descriptions)
 
         if summary := response.get("summary"):
             print("Updating PR description with summary...")
-            event.update_pr_description(number, f"{SUMMARY_MARKER}\n\n{ACTIONS_CREDIT}\n\n{summary}")
+            event.update_pr_description(number, f"{SUMMARY_MARKER}\n\n{ACTIONS_CREDIT}\n\n{summary}", pr_data.get("body"))
         else:
             summary = body
 
@@ -214,6 +217,9 @@ def main(*args, **kwargs):
         return
 
     # Handle issues and discussions (NOT PRs)
+    available_labels = event.get_repo_data("labels")
+    label_descriptions = {label["name"]: label.get("description") or "" for label in available_labels}
+
     current_labels = (
         []
         if issue_type == "discussion"
--- a/actions/first_interaction.py
+++ b/actions/first_interaction.py
@@ -195,7 +195,9 @@ def main(*args, **kwargs):
 
         if summary := response.get("summary"):
             print("Updating PR description with summary...")
-            event.update_pr_description(number, f"{SUMMARY_MARKER}\n\n{ACTIONS_CREDIT}\n\n{summary}", pr_data.get("body"))
+            event.update_pr_description(
+                number, f"{SUMMARY_MARKER}\n\n{ACTIONS_CREDIT}\n\n{summary}", pr_data.get("body")
+            )
         else:
             summary = body
 
--- a/actions/first_interaction.py
+++ b/actions/first_interaction.py
@@ -219,21 +219,22 @@ def main(*args, **kwargs):
         return
 
     # Handle issues and discussions (NOT PRs)
-    available_labels = event.get_repo_data("labels")
-    label_descriptions = {label["name"]: label.get("description") or "" for label in available_labels}
+    if issue_type != "pull request":
+        available_labels = event.get_repo_data("labels")
+        label_descriptions = {label["name"]: label.get("description") or "" for label in available_labels}
 
-    current_labels = (
-        []
-        if issue_type == "discussion"
-        else [label["name"].lower() for label in event.get_repo_data(f"issues/{number}/labels")]
-    )
+        current_labels = (
+            []
+            if issue_type == "discussion"
+            else [label["name"].lower() for label in event.get_repo_data(f"issues/{number}/labels")]
+        )
 
-    relevant_labels = get_relevant_labels(issue_type, title, body, label_descriptions, current_labels)
-    apply_and_check_labels(event, number, node_id, issue_type, username, relevant_labels, label_descriptions)
+        relevant_labels = get_relevant_labels(issue_type, title, body, label_descriptions, current_labels)
+        apply_and_check_labels(event, number, node_id, issue_type, username, relevant_labels, label_descriptions)
 
-    if action in {"opened", "created"}:
-        custom_response = get_first_interaction_response(event, issue_type, title, body, username)
-        event.add_comment(number, node_id, custom_response, issue_type)
+        if action in {"opened", "created"}:
+            custom_response = get_first_interaction_response(event, issue_type, title, body, username)
+            event.add_comment(number, node_id, custom_response, issue_type)
 
 
 if __name__ == "__main__":
--- a/actions/review_pr.py
+++ b/actions/review_pr.py
@@ -378,7 +378,9 @@ def dismiss_previous_reviews(event: Action) -> int:
                 event.put(f"{reviews_base}/{review_id}/dismissals", json={"message": "Superseded by new review"})
 
     # Batch delete all inline comments (with author verification)
-    comment_ids = [c["databaseId"] for c in comments if c.get("databaseId") and c.get("author", {}).get("login") == bot_username]
+    comment_ids = [
+        c["databaseId"] for c in comments if c.get("databaseId") and c.get("author", {}).get("login") == bot_username
+    ]
     event.batch_delete_review_comments(comment_ids)
 
     return review_count + 1
--- a/actions/review_pr.py
+++ b/actions/review_pr.py
@@ -361,11 +361,12 @@ def dismiss_previous_reviews(event: Action) -> int:
     if not (pr_number := event.pr.get("number")):
         return 1
 
-    # Fetch reviews and comments in single GraphQL query
-    reviews, comments = event.get_pr_reviews_and_comments(pr_number)
     bot_username = event.get_username()
     if not bot_username:
         return 1
+
+    # Fetch reviews and comments in single GraphQL query
+    reviews, comments = event.get_pr_reviews_and_comments(pr_number)
 
     review_count = 0
     reviews_base = f"{GITHUB_API_URL}/repos/{event.repository}/pulls/{pr_number}/reviews"
--- a/actions/review_pr.py
+++ b/actions/review_pr.py
@@ -358,29 +358,27 @@ def generate_pr_review(
 
 def dismiss_previous_reviews(event: Action) -> int:
     """Dismiss previous bot reviews and delete inline comments, returns count for numbering."""
-    if not (pr_number := event.pr.get("number")) or not (bot_username := event.get_username()):
+    if not (pr_number := event.pr.get("number")):
+        return 1
+
+    # Fetch reviews and comments in single GraphQL query
+    reviews, comments = event.get_pr_reviews_and_comments(pr_number)
+    bot_username = event.get_username()
+    if not bot_username:
         return 1
 
     review_count = 0
     reviews_base = f"{GITHUB_API_URL}/repos/{event.repository}/pulls/{pr_number}/reviews"
-    reviews_url = f"{reviews_base}?per_page=100"
-    if (response := event.get(reviews_url)).status_code == 200:
-        for review in response.json():
-            if review.get("user", {}).get("login") == bot_username and REVIEW_MARKER in (review.get("body") or ""):
-                review_count += 1
-                if review.get("state") in ["APPROVED", "CHANGES_REQUESTED"] and (review_id := review.get("id")):
-                    event.put(f"{reviews_base}/{review_id}/dismissals", json={"message": "Superseded by new review"})
+    for review in reviews:
+        # Double-check author matches bot (defense in depth)
+        if review.get("author", {}).get("login") == bot_username and REVIEW_MARKER in (review.get("body") or ""):
+            review_count += 1
+            if review.get("state") in ["APPROVED", "CHANGES_REQUESTED"] and (review_id := review.get("databaseId")):
+                event.put(f"{reviews_base}/{review_id}/dismissals", json={"message": "Superseded by new review"})
 
-    # Delete previous inline comments
-    comments_base = f"{GITHUB_API_URL}/repos/{event.repository}/pulls/{pr_number}/comments"
-    comments_url = f"{comments_base}?per_page=100"
-    if (response := event.get(comments_url)).status_code == 200:
-        for comment in response.json():
-            if comment.get("user", {}).get("login") == bot_username and (comment_id := comment.get("id")):
-                event.delete(
-                    f"{comments_base}/{comment_id}",
-                    expected_status=[200, 204, 404],
-                )
+    # Batch delete all inline comments (with author verification)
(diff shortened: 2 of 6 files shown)


actions/utils/github_utils.py
```diff
@@ -400,6 +450,61 @@ def handle_alert(self, number: int, node_id: str, issue_type: str, username: str
...
+        comments = pr_data.get("comments", {}).get("nodes", [])
+        return reviews, comments
+
+    def batch_delete_review_comments(self, comment_ids: list[str]) -> None:
+        """Batch delete PR review comments using a single GraphQL mutation."""
+        if not comment_ids:
+            return
+
+        mutations = []
+        for i, comment_id in enumerate(comment_ids):
```
UltralyticsAssistant: 💡 **MEDIUM**: `batch_delete_review_comments` builds a single GraphQL mutation with one field per comment ID. For PRs with many bot comments, this risks hitting GitHub GraphQL query size or complexity limits and failing the whole operation. It may be safer to chunk `comment_ids` into smaller batches (e.g., 20–50 IDs per mutation) and send multiple requests, handling failures per batch.

actions/review_pr.py
```diff
@@ -358,29 +358,27 @@ def generate_pr_review(
 
 def dismiss_previous_reviews(event: Action) -> int:
     """Dismiss previous bot reviews and delete inline comments, returns count for numbering."""
-    if not (pr_number := event.pr.get("number")) or not (bot_username := event.get_username()):
+    if not (pr_number := event.pr.get("number")):
```
UltralyticsAssistant: 📝 **LOW**: `dismiss_previous_reviews` and `get_pr_reviews_and_comments` both call `event.get_username()`, and the GraphQL query already filters reviews/comments by `botLogin`. You then re-fetch `bot_username` and re-check the author in the loop. The extra `get_username()` call is unnecessary, and the double-filtering adds a bit of complexity. You could simplify by obtaining `bot_username` once here and passing it into `get_pr_reviews_and_comments`, keeping a single source of truth for the login.

actions/utils/github_utils.py
```diff
@@ -97,6 +97,56 @@
...
 """
 
+GRAPHQL_PR_DETAILS = """
+query($owner: String!, $repo: String!, $number: Int!) {
+    repository(owner: $owner, name: $repo) {
+        pullRequest(number: $number) {
+            body
+        }
+        labels(first: 100) {
+            nodes {
```
UltralyticsAssistant: 📝 **LOW**: `GRAPHQL_PR_DETAILS` limits repository labels to `first: 100`. If a repository has more than 100 labels, anything beyond the first page will never be seen by the first-interaction flow, which is a behavior change from the previous REST-based approach. If you rely on labels beyond the first page (e.g., in large repos), consider adding pagination support or documenting this assumption.

actions/utils/github_utils.py
```diff
@@ -400,6 +450,61 @@ def handle_alert(self, number: int, node_id: str, issue_type: str, username: str
...
+        mutation = f"mutation {{ {' '.join(mutations)} }}"
+        self.graphql_request(mutation)
+
+    def get_multiple_issue_node_ids(self, issue_numbers: list[int]) -> dict[int, str]:
+        """Gets multiple issue node IDs in a single GraphQL query."""
+        if not issue_numbers:
+            return {}
+
+        owner, repo = self.repository.split("/")
+        queries = []
```
UltralyticsAssistant: 📝 **LOW**: `get_multiple_issue_node_ids` interpolates `owner`, `repo`, and issue numbers directly into the GraphQL query string. While these values are expected to be safe, this pattern makes escaping mistakes easier and can break if repository names ever contain quotes or special characters. For consistency and robustness with the rest of the file, consider switching to a parameterized GraphQL query that passes `owner`, `repo`, and `issueNumbers` as variables.

actions/summarize_pr.py
```diff
@@ -82,11 +82,27 @@ def label_fixed_issues(event, pr_summary):
...
+    issue_node_ids = event.get_multiple_issue_node_ids(issue_numbers)
+
+    # Get label IDs once
+    label_ids = event.get_label_ids(["fixed"])
+    if not label_ids:
+        print("Warning: 'fixed' label not found")
+        return pr_credit
+
+    # Batch label and comment all issues using GraphQL mutations
+    for issue_num, issue_node_id in issue_node_ids.items():
```
UltralyticsAssistant: 💭 **SUGGESTION**: `issue_num` is not used in this loop, which can be slightly confusing. Iterating directly over the values makes the intent clearer.

**Suggested change:**
```suggestion
    for issue_node_id in issue_node_ids.values():
```

UltralyticsAssistant (commented):
## 🔍 PR Review

Overall this PR is a solid refactor that reduces REST calls by leveraging GraphQL for PR details, reviews, comments, and issue labeling. The main opportunities for improvement are around robustness and clarity: avoid potential GraphQL limits in `batch_delete_review_comments`, consider how the 100-label GraphQL limit might affect large repositories, reduce duplicated `get_username` usage in the review dismissal path, tighten the GraphQL query construction for multiple issues, and clean up a minor unused loop variable in `label_fixed_issues`.

💬 Posted 5 inline comments

actions/first_interaction.py
```diff
@@ -214,6 +219,9 @@ def main(*args, **kwargs):
         return
 
     # Handle issues and discussions (NOT PRs)
+    available_labels = event.get_repo_data("labels")
```
UltralyticsAssistant: 💡 **MEDIUM**: These label-fetching lines run unconditionally, even though the comment above says they are for issues and discussions only. For newly opened PRs you already fetched labels via `get_pr_details_and_labels`, so this adds an extra REST call and may override the cached label descriptions. Consider guarding this block with `if issue_type != "pull request":` so it only runs for issues/discussions and avoids redundant API calls.

actions/review_pr.py
```diff
@@ -358,29 +358,28 @@ def generate_pr_review(
...
```
UltralyticsAssistant: 📝 **LOW**: `dismiss_previous_reviews` now calls `event.get_username()` directly and then calls `event.get_pr_reviews_and_comments(pr_number)`, which internally calls `get_username()` again. If `get_username()` performs a network lookup before caching, this is an unnecessary extra call. Consider passing `bot_username` into `get_pr_reviews_and_comments` or having that helper accept an optional `bot_login` argument to reuse the already-fetched value.

actions/review_pr.py
```diff
@@ -358,29 +358,28 @@ def generate_pr_review(
...
-                    expected_status=[200, 204, 404],
-                )
+    for review in reviews:
+        # Double-check author matches bot (defense in depth)
+        if review.get("author", {}).get("login") == bot_username and REVIEW_MARKER in (review.get("body") or ""):
+            review_count += 1
+            if review.get("state") in ["APPROVED", "CHANGES_REQUESTED"] and (review_id := review.get("databaseId")):
+                event.put(f"{reviews_base}/{review_id}/dismissals", json={"message": "Superseded by new review"})
+
+    # Batch delete all inline comments (with author verification)
```
UltralyticsAssistant: 📝 **LOW**: `batch_delete_review_comments` is invoked without any error handling or logging. If the GraphQL mutation fails (e.g., partial deletion, permission issues), the caller will silently assume comments were removed. It may be safer to either check the result of `graphql_request` or at least log failures for easier debugging.

UltralyticsAssistant (commented):
## 🔍 PR Review

Overall the PR is well-aligned with the goal of reducing REST calls by introducing targeted GraphQL helpers and reusing existing payload data. The main functional concern is that label fetching at the end of `first_interaction.main` now runs for all event types, including PRs, which both contradicts the comment and reintroduces an extra labels REST call on PR-open events. The rest of the changes are primarily structural and performance-oriented; I’ve noted a few minor clarity and robustness improvements around duplicate username lookups, silent failures in batch deletion, and the `update_pr_description` signature change potentially impacting external call sites.

💬 Posted 5 inline comments

UltralyticsAssistant (commented):
## 🔍 PR Review 3

Overall the refactor to use GraphQL for PR details, reviews, and issue labeling looks solid and should reduce REST chatter while keeping behavior consistent. The main improvement to address is ensuring `update_pr_description` still refreshes the PR body when the cached `current_body` is effectively missing (empty string), so you don’t accidentally drop existing descriptions. The remaining comments are minor optimizations around duplicate lookups, batching behavior, and avoiding unnecessary label fetches.

💬 Posted 5 inline comments

UltralyticsAssistant (commented):
## 🔍 PR Review

Overall the refactor to use GraphQL for PR details, reviews, and issue labeling looks solid and aligns with the goal of reducing REST calls. The main functional issue is that `get_pr_reviews_and_comments` reads from a `comments` field that doesn’t exist in the GraphQL response, so inline review comments are never deleted. Additionally, `batch_delete_review_comments` only handles the first 50 comments, which may leave older bot comments behind on large PRs. There’s also a minor efficiency concern around redundant `get_username` calls between `dismiss_previous_reviews` and `get_pr_reviews_and_comments`. Addressing these points will restore expected behavior while keeping the new performance benefits.

💬 Posted 3 inline comments

UltralyticsAssistant (commented):
## 🔍 PR Review 5

Overall the refactor meaningfully reduces REST calls by introducing GraphQL helpers for PR details, reviews/comments, and issue labeling, and the call-site changes in `first_interaction.py` and `summarize_pr.py` look consistent with the new APIs. The main areas to improve are: (1) hardening `update_pr_description` against failed/non-JSON GET responses, (2) making `batch_delete_review_comments` handle more than 50 comments or at least log when truncation occurs, (3) removing the now-unused `GITHUB_API_URL` import from `summarize_pr.py`, and (4) simplifying `dismiss_previous_reviews` to avoid redundant username lookups and filtering logic. The new GraphQL helpers themselves are generally well-structured and parameterized safely.

💬 Posted 4 inline comments

UltralyticsAssistant (commented):
## 🔍 PR Review 6

Overall the refactor to use GraphQL for PR metadata, reviews, and issue labeling looks solid and should reduce API calls. The main areas to tighten are around robustness and behavior parity with the old REST flow: ensure failures in the new GraphQL-based dismissal path are surfaced (or have a fallback), and restore tolerant handling of 404s when deleting old review comments. A bit more defensive error handling in `update_pr_description` and clear bounds for the batched issue ID lookup will make the new code more resilient.

💬 Posted 5 inline comments

UltralyticsAssistant (commented):
## 🔍 PR Review 6

Overall, the GraphQL batching changes are well-aligned with the goal of reducing REST calls and look consistent across PR review, summary, and issue labeling flows. The main functional concern is the robustness of the `GRAPHQL_PR_REVIEWS_COMMENTS` query: if the `author` filter is not supported as written, dismissing previous reviews will quietly stop working. Additionally, the signature change to `update_pr_description` may break external callers, and there is a small opportunity to avoid duplicate username resolution in the review dismissal path and to centralize label-fetching logic between PR and non-PR flows.

💬 Posted 5 inline comments

actions/utils/github_utils.py
```diff
@@ -273,27 +323,29 @@ def graphql_request(self, query: str, variables: dict | None = None) -> dict:
             print(result.get("errors"))
         return result
 
-    def update_pr_description(self, number: int, new_summary: str, max_retries: int = 2):
-        """Updates PR description with summary, retrying if description is None."""
+    def update_pr_description(self, number: int, new_summary: str, current_body: str | None = None):
```
UltralyticsAssistant, line 334: 💡 **MEDIUM**: `update_pr_description` changed signature from `(number, new_summary, max_retries=2)` to `(number, new_summary, current_body=None)` and hard-codes 3 attempts. Any existing callers that passed a positional third argument expecting `max_retries` will now treat that value as the PR body, which could cause incorrect description updates. Also, the maximum retries is now implicit and no longer configurable. It would be safer either to preserve the old parameter (e.g. add `max_retries` as a keyword-only argument) or to ensure all internal callers use keyword arguments so this change cannot break external usages.

actions/review_pr.py
```diff
@@ -358,29 +358,30 @@ def generate_pr_review(
 
 def dismiss_previous_reviews(event: Action) -> int:
     """Dismiss previous bot reviews and delete inline comments, returns count for numbering."""
-    if not (pr_number := event.pr.get("number")) or not (bot_username := event.get_username()):
+    if not (pr_number := event.pr.get("number")):
```
UltralyticsAssistant, line 380: 📝 **LOW**: `dismiss_previous_reviews` now calls `event.get_username()` directly and then indirectly again via `event.get_pr_reviews_and_comments()`. This double lookup is minor but unnecessary overhead on every run, and also duplicates the `None` handling logic in two places. Consider passing the already-resolved `bot_username` into `get_pr_reviews_and_comments` (or having it accept an optional precomputed login) to avoid redundant API calls and keep the responsibility for username resolution in one place.

actions/utils/github_utils.py
```diff
@@ -273,27 +323,29 @@ def graphql_request(self, query: str, variables: dict | None = None) -> dict:
...
         url = f"{GITHUB_API_URL}/repos/{self.repository}/pulls/{number}"
-        description = ""
-        for i in range(max_retries + 1):
-            description = self.get(url).json().get("body") or ""
-            if description:
-                break
-            if i < max_retries:
-                print("No current PR description found, retrying...")
-                time.sleep(1)
+        if current_body is None:
```
UltralyticsAssistant, line 339: 📝 **LOW**: Inside `update_pr_description`, the retry loop always assigns `current_body = self.get(url).json().get("body") or ""`, so by the time you reach the summary replacement/appending logic `current_body` is guaranteed to be a string. You can simplify and clarify the control flow by removing the `for i in range(3)` magic number and extracting the "fetch-with-retries" into a small helper or a clearly named constant, which makes the behavior easier to reason about and maintain.

actions/utils/github_utils.py
```diff
@@ -400,6 +452,71 @@ def handle_alert(self, number: int, node_id: str, issue_type: str, username: str
...
             self.block_user(username)
 
+    def get_pr_details_and_labels(self, pr_number: int) -> dict:
+        """Gets PR details and repository labels in a single GraphQL query."""
+        owner, repo = self.repository.split("/")
+        variables = {"owner": owner, "repo": repo, "number": pr_number}
+        result = self.graphql_request(GRAPHQL_PR_DETAILS, variables=variables)
+        return result.get("data", {}).get("repository", {})
+
+    def get_pr_reviews_and_comments(self, pr_number: int) -> tuple[list, list]:
```
UltralyticsAssistant, line 467: 📝 **LOW**: `get_pr_reviews_and_comments` calls `self.get_username()` internally, but `dismiss_previous_reviews` has already resolved and validated `bot_username` before calling this helper. This leads to redundant API calls and slightly divergent behavior between the helper and caller. Consider accepting `bot_login` as an optional argument and using the passed-in value when available to avoid duplicate lookups and keep the logic for handling a missing bot username centralized.

actions/utils/github_utils.py
```diff
@@ -400,6 +452,71 @@ def handle_alert(self, number: int, node_id: str, issue_type: str, username: str
...
+            comments.extend(review.get("comments", {}).get("nodes", []))
+
+        return reviews, comments
+
+    def batch_delete_review_comments(self, comment_ids: list[int]) -> None:
+        """Delete PR review comments using REST API (GraphQL batch mutations hit complexity limits)."""
+        for comment_id in comment_ids:
+            self.delete(f"{GITHUB_API_URL}/repos/{self.repository}/pulls/comments/{comment_id}")
+
+    def get_multiple_issue_node_ids(self, issue_numbers: list[int]) -> dict[int, str]:
```
UltralyticsAssistant, line 492: 📝 **LOW**: `get_multiple_issue_node_ids` dynamically builds a GraphQL query with one variable per issue. This is safe given the use of variables but can grow the query size linearly with `issue_numbers`. Since `closingIssuesReferences` is already capped upstream, this is probably fine, but it may be worth documenting or enforcing a reasonable upper bound on `issue_numbers` here to avoid accidentally constructing very large queries if the method is reused elsewhere.

UltralyticsAssistant (commented):
## 🔍 PR Review 8

Overall the refactor to use GraphQL for PR details, reviews, and issue labeling looks solid and aligns with the goal of reducing API calls. The main functional concern is the `update_pr_description` signature change, which could break any positional third-argument callers that previously relied on `max_retries`. There are also a couple of minor inefficiencies where `get_username()` is called twice for the same operation, and some opportunities to clarify retry behavior and bounds in the new helper methods.

💬 Posted 5 inline comments

glenn-jocher:
Closing this stale GraphQL optimization PR. The current diff is too broad for the benefit, and the review cleanup path now performs get_pr_reviews_and_comments() but discards the result and still uses the existing REST calls, so the core optimization claim is not true in the live diff. The smaller request-session pooling change from #590 covers the safe performance win.

PR #735, opened 2026-04-02
Dependabot action to consolidate PRs per dependency

glenn-jocher:
## 🛠️ PR Summary

### 🌟 Summary
This PR improves the custom Dependabot action by grouping related GitHub Action updates into fewer PRs, making version updates safer, cleaner, and easier to manage across repositories 🚀

### 📊 Key Changes
- 🔄 **Grouped updates into one PR per action version bump** instead of creating a separate PR for every file change.
- 📁 **Single PRs can now update multiple workflow files at once**, reducing PR noise and keeping related updates together.
- 🛡️ **Added safer version update logic** so specific tags are only upgraded when the latest version is in the same or a newer major release.
- 🏷️ **Simplified PR titles** to a consistent Dependabot-style format focused on `.github/workflows`.
- 📋 **Checks existing open PR titles before creating new ones**, helping avoid duplicate update PRs.
- 🧩 **Applies multiple replacements per file carefully in reverse order**, preventing broken edits when several updates happen in the same file.
- 📚 **Added a new `dependabot/README.md`** with setup instructions, inputs, behavior details, and caching explanation.

### 🎯 Purpose & Impact
- ✅ **Reduces PR clutter** by combining related workflow updates into fewer, more meaningful pull requests.
- 👀 **Makes dependency updates easier to review** since all references to the same action are updated together.
- 🔒 **Improves update safety** by avoiding inappropriate version jumps in some tagged-release scenarios.
- ⚡ **Better suited for organizations with many repos** thanks to grouped updates and cached version lookups.
- 📖 **Improves usability and adoption** with clear documentation for teams using this internal Dependabot alternative.
- 🏢 **Especially helpful for private or internal repos** where GitHub’s built-in Dependabot may not be available.

### DIFF
--- a/actions/dependabot.py
+++ b/actions/dependabot.py
@@ -81,10 +81,11 @@ def compute_update(current_ref, comment, latest):
         return None
 
     if re.fullmatch(r"v?\d+", current_ref):
-        # Major-only tag like @v6 -> update to latest major @v7
+        # Major-only tag like @v6 -> update to @v8 (or @v8.0.0 if v8 tag doesn't exist)
         if int(latest_major.group(1)) > int(current_major.group(1)):
             prefix = "v" if current_ref.startswith("v") else ""
-            return f"{prefix}{latest_major.group(1)}", comment
+            major_tag = f"{prefix}{latest_major.group(1)}"
+            return major_tag, comment
     elif current_ref != latest_tag:
         # Specific tag like @v2.8.0 -> update to latest tag
         return latest_tag, comment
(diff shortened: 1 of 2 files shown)


actions/dependabot.py
```diff
@@ -249,25 +263,42 @@ def run():
...
+            title = make_pr_title(action_repo, f"{new_ref}{info['new_comment']}")
+
+            if title in open_titles:
+                print(f"  ⏭️  {title} (PR already exists)")
+                total_prs_skipped += 1
+                continue
+
+            # Apply replacements per file (reverse order to preserve positions within each file)
+            file_updates = {}
+            for path, start, end, indent, action, nref, ncomment in info["replacements"]:
```
UltralyticsAssistant: ⚠️ **HIGH**: These replacements reuse `start`/`end` offsets captured from the original file content, but they are applied sequentially against the progressively modified string. If the same action appears more than once in a file and any earlier replacement changes the line length, later offsets will point at the wrong slice and can corrupt the file. Please group replacements by path and apply each file's edits in descending `start` order.

actions/dependabot.py
```diff
@@ -153,24 +164,23 @@ def create_single_pr(org, repo, title, file_path, new_content, token):
...
-            "branch": branch,
-        },
-    )
-    if r.status_code not in (200, 201):
-        print(f"    Failed to update {file_path}: {r.json().get('message', '')}")
-        return None
+    # Update each file on the branch
+    for path, content in file_updates:
+        r = requests.get(f"https://api.github.com/repos/{org}/{repo}/contents/{path}", headers=headers)
+        if r.status_code != 200:
```
UltralyticsAssistant, line 170: 💡 **MEDIUM**: If fetching or updating any file fails, this loop just skips that file and still proceeds to open the PR. That can create a 'successful' dependency bump PR that only updates a subset of the intended workflow files, which is especially hard to notice now that multiple files are grouped into one PR. It would be safer to abort PR creation (or surface a failure back to the caller) when any requested file update does not succeed.

UltralyticsAssistant (commented):
## 🔍 PR Review

Good refactor overall: the grouping flow and documentation make the action much easier to use. The main issue is that multi-file/multi-occurrence replacements currently apply original offsets against already-mutated content, which can break files when the same action appears multiple times. I also found one reliability issue where grouped PRs can be created with partial updates if any file write fails.

💬 Posted 2 inline comments

UltralyticsAssistant:
Huge thanks for merging this, @glenn-jocher — this is a fantastic improvement to the custom Dependabot workflow. By grouping related action updates, reducing PR noise, and adding safer version handling, this PR makes maintenance much cleaner and more scalable across repositories.

> “The secret of getting ahead is getting started.” — Mark Twain

This update is a great example of that mindset in action: thoughtful automation, clearer documentation, and safer defaults that will make life easier for every team relying on these workflows. Really appreciate the care that went into both the implementation and the developer experience here 🙌

PR #753, opened 2026-05-28
Switch loop guard to author-based to survive interleaving auto-commits

raimbekovm:
## summary

The "Commit and Push Changes" step in `action.yml` defends against recursive Auto-format loops by counting the last two commits whose message equals `Auto-format by https://ultralytics.com/actions`. The guard fires only when both top commits match that literal string, which fails to fire whenever a sibling auto-commit workflow interleaves a different message between two Auto-format pushes.

Observed live in ultralytics/ultralytics PR #24629: `.github/workflows/docs.yml` auto-commits `Auto-update Ultralytics Docs Reference by https://ultralytics.com/actions` on every PR push, and rebase merges produced by `pull_request: synchronize` pushes add `Merge ... into ...` commits. Both interleave between every Auto-format commit, so the message counter never reaches 2 and the loop ran 5+ cycles in 10 minutes before the PR was closed.

Replays of the actual `git log` show the message-based guard says RUN on all 9 sampled positions (loop survives); switching to `if [ "$(git log -1 --pretty=format:'%an')" = "$GITHUB_USERNAME" ]` says SKIP on every bot-authored HEAD and RUN only on the 3 legitimate human merge commits. The `GITHUB_USERNAME` env var is already set in this step (`env: GITHUB_USERNAME: ${{ inputs.github_username }}`), so the gate stays portable across orgs that configure a different bot user.

## 🛠️ PR Summary

### 🌟 Summary
🔁 This PR makes the auto-format GitHub Action smarter by preventing bot-triggered recursive commit loops while still formatting human-authored pull request updates as expected.

### 📊 Key Changes
- Added a new safety check to the **Commit and Push Changes** step in `action.yml`.
- The action now skips auto-committing when `github.actor` matches `inputs.github_username`, meaning the workflow was triggered by the same bot account.
- Replaced the older logic that checked whether the last two commit messages were both `"Auto-format by https://ultralytics.com/actions"`.
- Updated inline comments to clearly explain why bot-authored pushes can cause recursive workflow loops and why checking the GitHub actor is more reliable.

### 🎯 Purpose & Impact
- 🛑 **Prevents infinite or repeated auto-format commit loops** caused by bot-triggered pushes.
- ✅ **Improves reliability** by using the actual GitHub-triggering actor instead of guessing based on recent commit messages.
- 👤 **Preserves normal behavior for human contributors**, so PRs opened or updated by people still get auto-formatted.
- 🤝 **Reduces interference between related automation workflows**, such as formatting and docs update workflows that may use the same token.
- ⚡ **Makes CI behavior more predictable and efficient**, avoiding unnecessary reruns and extra bot commits.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -346,11 +346,11 @@ runs:
       run: |
         echo "::group::Commit and Push Changes"
         COMMIT_MSG="Auto-format by https://ultralytics.com/actions"
-        # Skip when HEAD was authored by the same bot to break recursive loops with sibling auto-commit workflows.
-        # Match name AND email so a contributor cannot spoof the bot identity by setting only --author name.
-        HEAD_AUTHOR=$(git log -1 --pretty=format:'%an|%ae' 2>/dev/null)
-        if [ "$HEAD_AUTHOR" = "$GITHUB_USERNAME|$GITHUB_EMAIL" ]; then
-          echo "Skipping commit: HEAD was authored by $GITHUB_USERNAME <$GITHUB_EMAIL>"
+        # Skip when this workflow run was triggered by the bot itself to break recursive loops with sibling
+        # auto-commit workflows (Auto-format <-> Auto-update Docs). GITHUB_ACTOR / GITHUB_TRIGGERING_ACTOR are
+        # set by GitHub from the auth token used to push and cannot be spoofed via local git author fields.
+        if [ "$GITHUB_ACTOR" = "$GITHUB_USERNAME" ] || [ "$GITHUB_TRIGGERING_ACTOR" = "$GITHUB_USERNAME" ]; then
+          echo "Skipping commit: workflow triggered by $GITHUB_USERNAME (sibling bot push)"
           echo "::endgroup::"
           exit 0
         fi
--- a/action.yml
+++ b/action.yml
@@ -346,11 +346,10 @@ runs:
       run: |
         echo "::group::Commit and Push Changes"
         COMMIT_MSG="Auto-format by https://ultralytics.com/actions"
-        # Check if last two commits were both auto-format to prevent recursive loops
-        LAST_TWO=$(git log -2 --pretty=format:'%s' 2>/dev/null || echo "")
-        MATCH_COUNT=$(printf '%s\n' "$LAST_TWO" | grep -Fxc "$COMMIT_MSG" || true)
-        if [ "$MATCH_COUNT" -ge 2 ]; then
-          echo "Skipping commit: last two commits were auto-format"
+        # Skip when HEAD was authored by the same bot to break recursive loops with sibling auto-commit workflows.
+        # Author-based instead of message-based so interleaving auto-commits (docs reference rebuilds, etc.) do not bypass the guard.
+        if [ "$(git log -1 --pretty=format:'%an' 2>/dev/null)" = "$GITHUB_USERNAME" ]; then
+          echo "Skipping commit: HEAD was authored by $GITHUB_USERNAME"
           echo "::endgroup::"
           exit 0
         fi
--- a/action.yml
+++ b/action.yml
@@ -343,14 +343,21 @@ runs:
       env:
         GITHUB_USERNAME: ${{ inputs.github_username }}
         GITHUB_EMAIL: ${{ inputs.github_email }}
+        GH_TOKEN: ${{ inputs.token }}
       run: |
         echo "::group::Commit and Push Changes"
         COMMIT_MSG="Auto-format by https://ultralytics.com/actions"
         # Skip when this workflow run was triggered by the bot itself to break recursive loops with sibling
-        # auto-commit workflows (Auto-format <-> Auto-update Docs). GITHUB_ACTOR / GITHUB_TRIGGERING_ACTOR are
-        # set by GitHub from the auth token used to push and cannot be spoofed via local git author fields.
-        if [ "$GITHUB_ACTOR" = "$GITHUB_USERNAME" ] || [ "$GITHUB_TRIGGERING_ACTOR" = "$GITHUB_USERNAME" ]; then
-          echo "Skipping commit: workflow triggered by $GITHUB_USERNAME (sibling bot push)"
+        # auto-commit workflows (Auto-format <-> Auto-update Docs). Derive the bot login from the token owner
+        # (single source of truth) so the guard works regardless of how callers configure github_username.
+        # GITHUB_ACTOR / GITHUB_TRIGGERING_ACTOR are set by GitHub from the push auth token and cannot be
+        # spoofed via local git author fields. Falls back to GITHUB_USERNAME if the API call fails.
+        BOT_LOGIN="$GITHUB_USERNAME"
+        if RESPONSE=$(gh api /user --jq '.login // empty' 2>/dev/null) && [ -n "$RESPONSE" ]; then
+          BOT_LOGIN="$RESPONSE"
+        fi
+        if [ "$GITHUB_ACTOR" = "$BOT_LOGIN" ] || [ "$GITHUB_TRIGGERING_ACTOR" = "$BOT_LOGIN" ]; then
+          echo "Skipping commit: workflow triggered by $BOT_LOGIN (sibling bot push)"
           echo "::endgroup::"
           exit 0
         fi
--- a/action.yml
+++ b/action.yml
@@ -339,23 +339,22 @@ runs:
 
     # Commit Changes ---------------------------------------------------------------------------------------------------
     - name: Commit and Push Changes
-      if: (github.event_name == 'pull_request' || github.event_name == 'pull_request_target') && (github.event.action == 'opened' || github.event.action == 'synchronize')
+      # Skip when the triggering push came from the bot itself (a prior Auto-format push, or a sibling
+      # auto-commit workflow sharing the same token such as Auto-update Docs). github.actor is set by GitHub
+      # from the pushing token and cannot be spoofed via git config, so this breaks recursive auto-commit loops
+      # even when sibling bot commits or merge commits interleave between Auto-format pushes. Human pushes have a
+      # different actor and are always formatted. Default-GITHUB_TOKEN pushes never re-trigger workflows, so only
+      # PAT-authored bot pushes (actor == github_username) can recurse here.
+      if: |
+        (github.event_name == 'pull_request' || github.event_name == 'pull_request_target') &&
+        (github.event.action == 'opened' || github.event.action == 'synchronize') &&
+        github.actor != inputs.github_username
       env:
         GITHUB_USERNAME: ${{ inputs.github_username }}
         GITHUB_EMAIL: ${{ inputs.github_email }}
       run: |
         echo "::group::Commit and Push Changes"
         COMMIT_MSG="Auto-format by https://ultralytics.com/actions"
-        # Skip when HEAD is this action's own auto-format commit. Narrow scope: only catches direct self-recursion
-        # of this step (e.g. manual re-run on a SHA we just pushed). Interleaving loops with sibling auto-commit
-        # workflows (e.g. Auto-update Docs) should be prevented at the caller's workflow `if:` level, since the
-        # set of sibling bots is consumer-specific.
-        HEAD_MSG=$(git log -1 --pretty=format:'%s' 2>/dev/null || echo "")
-        if [ "$HEAD_MSG" = "$COMMIT_MSG" ]; then
-          echo "Skipping commit: HEAD is already an auto-format commit"
-          echo "::endgroup::"
-          exit 0
-        fi
         git config --global user.name "$GITHUB_USERNAME"
         git config --global user.email "$GITHUB_EMAIL"
         git add .
--- a/action.yml
+++ b/action.yml
@@ -343,21 +343,16 @@ runs:
       env:
         GITHUB_USERNAME: ${{ inputs.github_username }}
         GITHUB_EMAIL: ${{ inputs.github_email }}
-        GH_TOKEN: ${{ inputs.token }}
       run: |
         echo "::group::Commit and Push Changes"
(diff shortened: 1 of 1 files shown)


UltralyticsAssistant:
👋 Hello @raimbekovm, thank you for submitting an automated `ultralytics/actions` 🚀 PR! This message is automatically generated to help streamline review, and an engineer will assist you shortly.

To ensure a seamless integration of your work, please review the following checklist:
- ✅ **Ensure CI Checks Pass**: Verify all Ultralytics [Continuous Integration (CI)](https://docs.ultralytics.com/help/CI/) checks are passing. If any checks fail, please address the issues.
- ✅ **Sign the CLA**: Please ensure you have signed our [Contributor License Agreement](https://docs.ultralytics.com/help/CLA/) if this is your first Ultralytics PR by writing "I have read the CLA Document and I sign the CLA" in a new message.
- ✅ **Minimize Changes**: Limit your changes to the **minimum** necessary for your bug fix or feature addition. _"It is not daily increase but daily decrease, hack away the unessential. The closer to the source, the less wastage there is."_  — Bruce Lee

For more guidance, please refer to our [Contributing Guide](https://docs.ultralytics.com/help/contributing/). Don't hesitate to leave a comment if you have any questions. Thank you for contributing to Ultralytics! 🚀🤖

UltralyticsAssistant (commented):
## 🔍 PR Review

The change is a reasonable simplification, but the new guard is too weak as implemented: checking only the latest commit's author name can be spoofed and may skip legitimate runs. Tightening the identity check would make this update safe.

💬 Posted 1 inline comment

action.yml
```diff
@@ -346,11 +346,10 @@ runs:
...
         echo "::group::Commit and Push Changes"
         COMMIT_MSG="Auto-format by https://ultralytics.com/actions"
-        # Check if last two commits were both auto-format to prevent recursive loops
-        LAST_TWO=$(git log -2 --pretty=format:'%s' 2>/dev/null || echo "")
-        MATCH_COUNT=$(printf '%s\n' "$LAST_TWO" | grep -Fxc "$COMMIT_MSG" || true)
-        if [ "$MATCH_COUNT" -ge 2 ]; then
-          echo "Skipping commit: last two commits were auto-format"
+        # Skip when HEAD was authored by the same bot to break recursive loops with sibling auto-commit workflows.
+        # Author-based instead of message-based so interleaving auto-commits (docs reference rebuilds, etc.) do not bypass the guard.
+        if [ "$(git log -1 --pretty=format:'%an' 2>/dev/null)" = "$GITHUB_USERNAME" ]; then
```
raimbekovm: Reproduced the spoof empirically: `git commit --author "UltralyticsAssistant <evil@example.com>"` lands a commit with `%an = UltralyticsAssistant` and the previous guard would skip the next Auto-format run. Tightened to match both name and email (`%an|%ae` against `$GITHUB_USERNAME|$GITHUB_EMAIL`) in b457182. Re-replayed against the original loop on PR #24629 — 6/6 bot-authored HEAD positions still SKIP (loop stays broken) and the spoofed `evil@example.com` commit now correctly RUNs.

UltralyticsAssistant (commented):
## 🔍 PR Review 2

The author-based loop guard is a good attempt to handle interleaved bot commits, but the current implementation trusts spoofable commit metadata and can be bypassed by any contributor. I’d fix that before merging; otherwise the diff looks focused and minimal.

💬 Posted 1 inline comment

pderrenger:
Thanks for tracking this down—using the latest commit author as the guard is a much more reliable signal than message matching here and should stop the recursive bot-commit churn even when other automated commits get interleaved.

action.yml
```diff
@@ -346,11 +346,11 @@ runs:
...
         COMMIT_MSG="Auto-format by https://ultralytics.com/actions"
-        # Check if last two commits were both auto-format to prevent recursive loops
-        LAST_TWO=$(git log -2 --pretty=format:'%s' 2>/dev/null || echo "")
-        MATCH_COUNT=$(printf '%s\n' "$LAST_TWO" | grep -Fxc "$COMMIT_MSG" || true)
-        if [ "$MATCH_COUNT" -ge 2 ]; then
-          echo "Skipping commit: last two commits were auto-format"
+        # Skip when HEAD was authored by the same bot to break recursive loops with sibling auto-commit workflows.
+        # Match name AND email so a contributor cannot spoof the bot identity by setting only --author name.
+        HEAD_AUTHOR=$(git log -1 --pretty=format:'%an|%ae' 2>/dev/null)
+        if [ "$HEAD_AUTHOR" = "$GITHUB_USERNAME|$GITHUB_EMAIL" ]; then
```
raimbekovm: Confirmed empirically — both `%an` and `%ae` are fully user-controlled. Reproduced with a single command:

```
$ git commit --author "UltralyticsAssistant <web@ultralytics.com>" -m "looks legit"
$ git log -1 --pretty=format:'%an|%ae'
UltralyticsAssistant|web@ultralytics.com   ← matches the guard, formatter silently skipped
```

Switched to `GITHUB_ACTOR` / `GITHUB_TRIGGERING_ACTOR`, which GitHub sets from the auth token used to push and which cannot be set from local git config. This is the same signal sibling workflows already use at the `if:` level (`github.actor != 'UltralyticsAssistant'`), now applied inside the action so any consumer gets the protection without duplicating it in their workflow.

Verified across 4 contexts (bot push → SKIP, spoofed-author push → RUN, manual re-run by maintainer → SKIP, normal contributor push → RUN).

UltralyticsAssistant (commented):
## 🔍 PR Review 3

Good direction overall—the actor-based guard addresses the interleaving-commit loophole more directly than the old message counter. The main thing to fix is that the comparison is currently tied to the configurable commit author rather than the actual authenticated bot login, which makes the new guard fragile.

💬 Posted 1 inline comment

action.yml
```diff
@@ -346,11 +346,11 @@ runs:
...
         COMMIT_MSG="Auto-format by https://ultralytics.com/actions"
-        # Check if last two commits were both auto-format to prevent recursive loops
-        LAST_TWO=$(git log -2 --pretty=format:'%s' 2>/dev/null || echo "")
-        MATCH_COUNT=$(printf '%s\n' "$LAST_TWO" | grep -Fxc "$COMMIT_MSG" || true)
-        if [ "$MATCH_COUNT" -ge 2 ]; then
-          echo "Skipping commit: last two commits were auto-format"
+        # Skip when this workflow run was triggered by the bot itself to break recursive loops with sibling
+        # auto-commit workflows (Auto-format <-> Auto-update Docs). GITHUB_ACTOR / GITHUB_TRIGGERING_ACTOR are
+        # set by GitHub from the auth token used to push and cannot be spoofed via local git author fields.
+        if [ "$GITHUB_ACTOR" = "$GITHUB_USERNAME" ] || [ "$GITHUB_TRIGGERING_ACTOR" = "$GITHUB_USERNAME" ]; then
```
raimbekovm: Fair point — `github_username` is the git commit author display name input, separate in principle from the GitHub login behind `inputs.token`. They happen to align in the default setup but the action does not enforce that.

Switched to deriving the bot login from the token owner directly:

```bash
BOT_LOGIN="$GITHUB_USERNAME"
if RESPONSE=$(gh api /user --jq '.login // empty' 2>/dev/null) && [ -n "$RESPONSE" ]; then
  BOT_LOGIN="$RESPONSE"
fi
```

`gh api /user` resolves the authenticated user behind the token, so `BOT_LOGIN` always matches what `GITHUB_ACTOR` will be when the same token pushes. Falls back to `GITHUB_USERNAME` if the API call fails (no token, no network, no `gh`), preserving the previous behaviour for setups where they match by convention.

Verified the fallback path: an invalid token writes the error body to stdout but exits non-zero, so the `if` chain correctly takes the fallback branch. Also verified the divergent-name case (`github_username="Auto Format Bot"`, token owner `my-bot-account`): old guard fails to skip, new guard skips correctly.

UltralyticsAssistant (commented):
## 🔍 PR Review 4

Good direction on fixing the interleaving auto-commit loop, but the current actor-based guard is broader than the original intent. As written, it can suppress legitimate formatting runs when the token owner is not a dedicated bot and when sibling automation pushes changes that still need this action's fixes.

💬 Posted 2 inline comments

action.yml
```diff
@@ -343,14 +343,21 @@ runs:
...
-        MATCH_COUNT=$(printf '%s\n' "$LAST_TWO" | grep -Fxc "$COMMIT_MSG" || true)
-        if [ "$MATCH_COUNT" -ge 2 ]; then
-          echo "Skipping commit: last two commits were auto-format"
+        # Skip when this workflow run was triggered by the bot itself to break recursive loops with sibling
+        # auto-commit workflows (Auto-format <-> Auto-update Docs). Derive the bot login from the token owner
+        # (single source of truth) so the guard works regardless of how callers configure github_username.
+        # GITHUB_ACTOR / GITHUB_TRIGGERING_ACTOR are set by GitHub from the push auth token and cannot be
+        # spoofed via local git author fields. Falls back to GITHUB_USERNAME if the API call fails.
+        BOT_LOGIN="$GITHUB_USERNAME"
+        if RESPONSE=$(gh api /user --jq '.login // empty' 2>/dev/null) && [ -n "$RESPONSE" ]; then
```
raimbekovm: Confirmed — for setups passing a maintainer PAT, `gh api /user` resolves to that human's login, and every regular push of theirs would then match the guard and silently skip formatting. The actor-based check was correct for "is this loop our own bot?" but wrong for "should this PR get formatted?".

Reverted to a narrow message check on HEAD only: skip iff the HEAD commit message is exactly `Auto-format by https://ultralytics.com/actions`. That catches direct self-recursion of this step (manual re-run on a SHA we just pushed) without making any assumptions about who owns the token. A maintainer's regular commits don't match and get formatted normally.

action.yml
```diff
@@ -343,14 +343,21 @@ runs:
...
+          BOT_LOGIN="$RESPONSE"
+        fi
+        if [ "$GITHUB_ACTOR" = "$BOT_LOGIN" ] || [ "$GITHUB_TRIGGERING_ACTOR" = "$BOT_LOGIN" ]; then
```
raimbekovm: Right — the broader fix was suppressing formattable changes from any sibling bot run, not just our own recursion. With the message-based revert, only HEAD commits with this action's own message (`Auto-format by https://ultralytics.com/actions`) skip. A sibling workflow pushing `Auto-update Ultralytics Docs Reference by ...` no longer matches, so the follow-up run will format any unformatted changes that bot introduced.

Interleaving loops between this action and sibling auto-commit workflows are consumer-specific (the set of sibling bots varies by deployment), so they belong at the caller's workflow `if:` level — where the consumer can name which bot actors to skip — rather than inside this action. `ultralytics/format.yml` already does this at line 24.

action.yml
```diff
@@ -346,11 +346,13 @@ runs:
...
-        LAST_TWO=$(git log -2 --pretty=format:'%s' 2>/dev/null || echo "")
-        MATCH_COUNT=$(printf '%s\n' "$LAST_TWO" | grep -Fxc "$COMMIT_MSG" || true)
-        if [ "$MATCH_COUNT" -ge 2 ]; then
-          echo "Skipping commit: last two commits were auto-format"
+        # Skip when HEAD is this action's own auto-format commit. Narrow scope: only catches direct self-recursion
+        # of this step (e.g. manual re-run on a SHA we just pushed). Interleaving loops with sibling auto-commit
+        # workflows (e.g. Auto-update Docs) should be prevented at the caller's workflow `if:` level, since the
+        # set of sibling bots is consumer-specific.
+        HEAD_MSG=$(git log -1 --pretty=format:'%s' 2>/dev/null || echo "")
+        if [ "$HEAD_MSG" = "$COMMIT_MSG" ]; then
```
UltralyticsAssistant: ⚠️ **HIGH**: This guard no longer addresses the failure mode described in the PR. It only skips when the immediate `HEAD` commit message is the auto-format message, so an interleaved bot commit or merge commit will still let this step run again and recreate the loop. In other words, the change weakens the protection to direct self-recursion instead of fixing the interleaving case the PR is targeting.

UltralyticsAssistant (commented):
## 🔍 PR Review 5

The change is small and well-commented, but the new guard appears to miss the core loop scenario from the PR description. As written, it only blocks direct self-recursion on an auto-format `HEAD`, not interleaved bot/merge commits, so I’d address that before merging.

💬 Posted 1 inline comment

glenn-jocher (approved):
Approving the reworked fix.

The guard now gates the Commit and Push step at the if: level with `github.actor != inputs.github_username` and drops the bash message-matching block entirely.

Why this resolves the prior review thread:
- Non-spoofable: github.actor is set by GitHub from the pushing token and cannot be set via local git config, unlike the earlier %an/%ae author check.
- Fixes the interleaving loop: docs.yml and rebase-merge commits push with the same org bot token, so every bot-triggered synchronize has actor == github_username and is skipped regardless of commit message. Message-matching could not see past interleaved sibling/merge commits.
- No regression: human and contributor pushes have a different actor and are still formatted. Default GITHUB_TOKEN pushes do not re-trigger workflows, so only PAT-authored bot pushes can recurse here.
- Consistent with existing code: the PR Review step already compares github.event.requested_reviewer.login == inputs.github_username, treating github_username as the bot login.

All CI green on 5bcf92a.

UltralyticsAssistant:
Merged! 🎉

Big thanks to @raimbekovm for this thoughtful improvement, with contributions from @pderrenger and @glenn-jocher. This update makes the auto-format action much more robust by preventing bot-triggered recursive commit loops while preserving the expected experience for human contributors.

As Benjamin Franklin said, “An ounce of prevention is worth a pound of cure.” That fits this PR perfectly: a small, smart safeguard that improves CI reliability, reduces unnecessary reruns, and keeps automation predictable and efficient.

Really appreciate the careful attention to workflow stability here.

PR #851, opened 2026-07-30
Converse in review threads instead of deleting bot comments

JaviChulvi:
## Summary

The PR review bot currently deletes all of its previous inline comments before each new review and never sees replies, so it re-raises findings the author already rebutted — every round is a cold re-review and thread pushback can never converge.

**Motivation:** this fix is inspired by ultralytics/assistant#2978, where the bot kept bringing up the same finding round after round even after the author pushed back explaining the issue was not introduced by that PR and not considered important — because each new review deleted the previous comments and never read the replies, the pushback could never register and the only way to silence the bot was changing the code.

This PR makes review threads persistent and conversational:

- **Stop deleting inline comments.** `clear_previous_review` becomes `dismiss_previous_reviews`: it still dismisses stale APPROVED/CHANGES_REQUESTED states (so an outdated approval never counts for new commits) but the comment-deletion loop is removed.
- **Feed existing threads to the model.** `get_review_threads` fetches the bot's unresolved threads via paginated GraphQL `reviewThreads`, scoped to reviews carrying the bot's `REVIEW_MARKER` (threads from other tools sharing the `github-actions[bot]` login are ignored). Threads are rendered into the review prompt keyed `T1/T2/...` — root finding plus latest replies, whole threads only. Injected into the prompt (not a tool) so it also works on the Anthropic single-shot fallback, which strips function tools.
- **Per-thread verbs.** The model returns `thread_actions`: `resolve` (code now fixes it, or the rebuttal holds on substance) or `reply` (finding stands — answer posted in-thread). Applied via the REST replies endpoint and the `resolveReviewThread` mutation, before the summary posts so the review only ever claims actions that actually happened.

## Guardrails

- **Bounded debates:** two replies per thread is the limit — if the author still disagrees after the bot's second reply, the bot must resolve with a recorded dissent note or stand aside (enforced in code, not just prompt). No stalemate can block a PR forever.
- **Settled memory:** resolved bot threads are fed back as one-line "settled findings — do not re-raise" context, so conceded findings never return, on any line.
- **Severity calibration:** CRITICAL/HIGH require wrong behavior, data loss, security exposure, or wrong merge decisions; LOW/SUGGESTION threads persist as advice but do not block APPROVE.
- **Anchor dedup with escalation:** new comments on a line owned by an existing thread are filtered unless strictly more severe — no duplicate nagging, but a serious regression can never hide behind an advisory nit.
- Replies are suppressed when the bot already has the last word (no nag loops); rebuttals are accepted only on technical substance, and a reply that keeps a finding alive must cite the current code sustaining it.
- Actions are validated against the threads actually shown to the model; thread mutations are guarded against stale heads and conversations that changed during generation.
- Thread fetch and resolve fail loudly (missing username, GraphQL errors, rate-limit bodies) instead of degrading to "no threads".
- APPROVE is blocked while unresolved MEDIUM+ bot threads remain; the summary shows a 🔁 hint telling authors how to continue the conversation.

## Seen live on this PR

The whole loop ran on this PR itself (`format.yml` dogfoods the PR head): 9 findings across 11 review rounds — 5 fixed, 4 rebutted in-thread with every rebuttal evaluated and accepted, nothing resolved ever re-raised. When one truncation debate stalled after three exchanges, the debate limit ended it exactly as designed, with the reviewer recording its dissent and resolving its own thread: https://github.com/ultralytics/actions/pull/851#discussion_r3684391448 — and once every thread was fixed, conceded, or dissent-closed, the bot issued its APPROVE.

## Notes

- Threads fetch the root finding plus the `last: 30` replies so the newest rebuttal is always visible to the model and the last-word guard; bodies over 1,000 chars carry an explicit truncation marker.
- Uses `fullDatabaseId` (not the 32-bit `databaseId`, which modern comment IDs overflow) for reply targets.
- The author can get a response without pushing by re-requesting review from the bot.
- Existing test updated to assert comments are never deleted; `review_pr.py` is coverage-omitted, validation done via mocked-network smoke tests of all new paths (pagination, marker scoping, escalation, debate limits, failure modes).
- Bumps `__version__` to 0.2.45.

### DIFF
--- a/actions/review_pr.py
+++ b/actions/review_pr.py
@@ -686,7 +686,11 @@ def generate_pr_review(
         print(f"AI generated {comments_before_filtering} comments")
 
         # Validate, filter, and deduplicate comments
-        thread_anchors = {(t["path"], t["side"], t["line"]) for t in threads.values() if t.get("line")}
+        thread_anchors = {
+            (t["path"], t["side"], t["line"]): SEVERITY_RANK.get(t.get("severity") or "CRITICAL", 0)
+            for t in threads.values()
+            if t.get("line")
+        }
         unique_comments = {}
         for c in response.get("comments", []):
             file_path, line_num = c.get("file"), c.get("line", 0)
@@ -706,8 +710,9 @@ def generate_pr_review(
                 print(f"Filtered out {file_path}:{line_num} (side={side}, available: {available})")
                 continue
 
-            # An existing thread already anchors here; the model must use a thread action, not a duplicate comment
-            if (file_path, side, line_num) in thread_anchors:
+            # An existing thread anchors here: only a strictly more severe new finding may share the anchor
+            anchor_rank = thread_anchors.get((file_path, side, line_num))
+            if anchor_rank is not None and SEVERITY_RANK.get(c.get("severity"), 5) >= anchor_rank:
                 print(f"Filtered out {file_path}:{line_num} (existing review thread anchors here)")
                 continue
 
@@ -779,7 +784,7 @@ def generate_pr_review(
                 continue
             thread_actions[ref] = {"thread": ref, "action": verb, "message": message}
         # Only findings-level threads gate approval: LOW/SUGGESTION threads advise without blocking
-        blocking = {ref for ref, t in threads.items() if t.get("blocking", True)}
+        blocking = {ref for ref, t in threads.items() if t.get("severity") not in ("LOW", "SUGGESTION")}
         resolved = {ref for ref, a in thread_actions.items() if a["action"] == "resolve"}
 
         response.update(
@@ -886,7 +891,7 @@ def get_review_threads(event: Action) -> dict[str, dict]:
                 "line": node.get("line"),
                 "side": node.get("diffSide") or "RIGHT",
                 "outdated": bool(node.get("isOutdated")),
-                "blocking": not severity or severity.group(1) not in ("LOW", "SUGGESTION"),
+                "severity": severity.group(1) if severity else None,
                 "comments": [
                     {"author": (c.get("author") or {}).get("login") or "unknown", "body": c.get("body") or ""}
                     for c in comments
@@ -912,7 +917,7 @@ def apply_thread_actions(event: Action, review_data: dict, threads: dict[str, di
         # New replies don't change the head SHA: never act on a conversation the model didn't see
         if fresh.get(thread["id"]) != thread["comments"]:
             print(f"Skipping {action['action']} on thread {thread['id']}: conversation changed during review")
-            if action["action"] == "resolve" and thread.get("blocking", True):
+            if action["action"] == "resolve" and thread.get("severity") not in ("LOW", "SUGGESTION"):
                 review_data["open_threads"] += 1  # the thread stays open: keep the APPROVE gate honest
             continue
         if (message := action.get("message")) and (root_id := thread.get("root_comment_id")):
--- a/actions/review_pr.py
+++ b/actions/review_pr.py
@@ -49,12 +49,19 @@ query($owner: String!, $name: String!, $number: Int!, $cursor: String) {
                     diffSide
                     path
                     line
-                    comments(first: 30) {
+                    root: comments(first: 1) {
                         nodes {
                             fullDatabaseId
                             author { login }
                             body
                             pullRequestReview { body }
+                        }
+                    }
+                    latest: comments(last: 30) {
+                        nodes {
+                            fullDatabaseId
+                            author { login }
+                            body
                         }
                     }
                 }
@@ -850,14 +857,16 @@ def get_review_threads(event: Action) -> dict[str, dict]:
             "reviewThreads"
         ) or {}
         for node in connection.get("nodes") or []:
-            comments = (node.get("comments") or {}).get("nodes") or []
-            root = comments[0] if comments else {}
+            root = ((node.get("root") or {}).get("nodes") or [{}])[0]
             if (
                 node.get("isResolved")
                 or (root.get("author") or {}).get("login") != bot_username
                 or REVIEW_MARKER not in ((root.get("pullRequestReview") or {}).get("body") or "")
             ):
                 continue
+            comments = (node.get("latest") or {}).get("nodes") or []  # newest replies: the last-word guard needs them
+            if not comments or comments[0].get("fullDatabaseId") != root.get("fullDatabaseId"):
+                comments = [root, *comments]  # thread longer than the fetched tail: keep the root finding visible
             threads[f"T{len(threads) + 1}"] = {
                 "id": node["id"],
                 "root_comment_id": root.get("fullDatabaseId"),
@@ -878,8 +887,12 @@ def get_review_threads(event: Action) -> dict[str, dict]:
 
 def apply_thread_actions(event: Action, review_data: dict, threads: dict[str, dict]) -> None:
     """Reply to and resolve existing review threads as decided by the review model."""
+    if not (actions := review_data.get("thread_actions")):
+        return
+    if event.get_pr_head_sha() != review_data["head_sha"]:  # don't mutate threads from a stale review
+        raise RuntimeError("PR head changed during review generation")
     pr_number = event.pr["number"]
-    for action in review_data.get("thread_actions", []):
+    for action in actions:
         thread = threads[action["thread"]]
         if (message := action.get("message")) and (root_id := thread.get("root_comment_id")):
             event.post(
--- a/actions/review_pr.py
+++ b/actions/review_pr.py
@@ -893,6 +893,7 @@ def apply_thread_actions(event: Action, review_data: dict, threads: dict[str, di
         raise RuntimeError("PR head changed during review generation")
     fresh = {t["id"]: t["comments"] for t in get_review_threads(event).values()}
     pr_number = event.pr["number"]
+    applied = []
     for action in actions:
         thread = threads[action["thread"]]
         # New replies don't change the head SHA: never act on a conversation the model didn't see
@@ -911,6 +912,8 @@ def apply_thread_actions(event: Action, review_data: dict, threads: dict[str, di
             result = event.graphql_request(GRAPHQL_RESOLVE_REVIEW_THREAD, variables={"threadId": thread["id"]})
             if "data" not in result or result.get("errors"):
                 raise RuntimeError(f"Failed to resolve review thread {thread['id']}: {result}")
+        applied.append(action)
+    review_data["thread_actions"] = applied  # the summary's thread count reports only what actually happened
 
 
 def post_review_summary(event: Action, review_data: dict) -> None:
--- a/actions/review_pr.py
+++ b/actions/review_pr.py
@@ -34,6 +34,43 @@ MAX_TOOL_FILE_LINES = 240
 MAX_AGENT_TURNS = 16
 REVIEW_COST_SOFT_LIMIT = 2.00  # stop requesting tools after cumulative spend reaches this amount
 SEVERITY_RANK = {"CRITICAL": 0, "HIGH": 1, "MEDIUM": 2, "LOW": 3, "SUGGESTION": 4, None: 5}
+MAX_THREADS_SECTION_CHARS = 8000
+
+GRAPHQL_PR_REVIEW_THREADS = """
+query($owner: String!, $name: String!, $number: Int!, $cursor: String) {
+    repository(owner: $owner, name: $name) {
+        pullRequest(number: $number) {
+            reviewThreads(first: 100, after: $cursor) {
+                pageInfo { hasNextPage endCursor }
+                nodes {
+                    id
+                    isResolved
+                    isOutdated
+                    diffSide
+                    path
+                    line
+                    comments(first: 30) {
+                        nodes {
+                            fullDatabaseId
+                            author { login }
+                            body
+                            pullRequestReview { body }
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
+"""
+
+GRAPHQL_RESOLVE_REVIEW_THREAD = """
+mutation($threadId: ID!) {
+    resolveReviewThread(input: {threadId: $threadId}) {
+        thread { id }
+    }
+}
+"""
 
 
 def _clip_tool_output(text: str, limit: int = MAX_TOOL_OUTPUT_CHARS) -> str:
@@ -361,6 +398,32 @@ def parse_diff_files(diff_text: str) -> tuple[dict, str]:
     return files, "\n".join(augmented_lines)
 
 
+def format_review_threads(threads: dict[str, dict]) -> tuple[str, set[str]]:
+    """Format unresolved review threads as a bounded prompt section, returning the refs actually shown."""
+    blocks, shown, total = [], set(), 0
+    for ref, thread in threads.items():
+        anchor = f"{thread['path']}:{thread['line']} ({thread['side']})" if thread.get("line") else thread["path"]
+        outdated = " [outdated: the anchored code changed since]" if thread["outdated"] else ""
+        comments = thread["comments"]
+        if len(comments) > 10:  # root finding plus the latest replies: the newest rebuttal matters most
+            comments = [comments[0], *comments[-9:]]
+        convo = "\n".join(f"  {c['author']}: {c['body'][:1000]}" for c in comments)
+        block = _clip_tool_output(f"[{ref}] {anchor}{outdated}\n{convo}", MAX_THREADS_SECTION_CHARS)
+        if total + len(block) > MAX_THREADS_SECTION_CHARS and blocks:  # whole threads only, never cut mid-thread
+            print(f"Dropping {len(threads) - len(shown)} review threads from prompt (section budget)")
+            break
+        blocks.append(block)
+        shown.add(ref)
+        total += len(block) + 2
+    if not blocks:
+        return "", shown
+    section = "\n\n".join(blocks)
+    return (
+        f"EXISTING REVIEW THREADS (your unresolved findings from previous reviews, with replies):\n{section}\n\n",
+        shown,
+    )
+
+
 def generate_pr_review(
     repository: str,
     diff_text: str,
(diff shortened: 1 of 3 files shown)


UltralyticsAssistant:
👋 Hello @JaviChulvi, thank you for submitting a `ultralytics/actions` 🚀 PR! This review is automated, and an engineer will assist with any follow-up. To ensure a seamless integration of your work, please review the following checklist:

UltralyticsAssistant (commented):
## 🔍 PR Review

The persistent-thread design and marker-based ownership filtering are solid, but the implementation can miss the newest conversation replies and can mutate thread state after the reviewed commit has become stale.

💬 Posted 2 inline comments

actions/review_pr.py
```diff
@@ -804,10 +996,14 @@ def main(*args, **kwargs):
...
-    clear_previous_review(event)
+    dismiss_previous_reviews(event)
+    threads = get_review_threads(event)
+    if threads:
+        print(f"Found {len(threads)} unresolved review threads from previous reviews")
     review = generate_pr_review(
-        event.repository, diff, event.pr.get("title") or "", event.pr.get("body") or "", event, head_sha
+        event.repository, diff, event.pr.get("title") or "", event.pr.get("body") or "", event, head_sha, threads
     )
+    apply_thread_actions(event, review, threads)  # before the summary, which claims these actions and may APPROVE
```
UltralyticsAssistant, line 1083: ⚠️ **HIGH**: `apply_thread_actions` mutates existing threads before `post_review_summary()` performs its existing head-SHA validation. If a commit is pushed while generation is running, the summary correctly aborts as stale, but replies and resolutions have already been applied using the old diff and conversation. Validate that the current head still equals `review["head_sha"]` before applying these actions, and abort/retry on mismatch.

actions/review_pr.py
```diff
@@ -34,6 +34,43 @@
...
+            reviewThreads(first: 100, after: $cursor) {
+                pageInfo { hasNextPage endCursor }
+                nodes {
+                    id
+                    isResolved
+                    isOutdated
+                    diffSide
+                    path
+                    line
+                    comments(first: 30) {
```
UltralyticsAssistant: 💡 **MEDIUM**: This fetches only the first 30 comments in a thread, while the later formatting and reply guard treat the end of this list as the latest reply. For a thread with more than 30 comments, newer author rebuttals or bot replies are omitted, so the bot can re-raise a rebutted finding or post a duplicate response—the behavior this change is intended to prevent. Paginate the nested `comments` connection before constructing the thread data. GitHub GraphQL connections are paginated and support `first`/`after` traversal.

actions/review_pr.py
UltralyticsAssistant: `apply_thread_actions()` now validates the live PR head against `review_data['head_sha']` before posting replies or resolving threads.

actions/review_pr.py
UltralyticsAssistant: The query now uses `comments(last: 30)`, so the newest replies are retained for the last-word guard and prompt formatting; this addresses the prior first-page truncation concern.

actions/review_pr.py
```diff
@@ -804,10 +1009,14 @@ def main(*args, **kwargs):
     except RuntimeError as e:
         print(f"Skipping stale PR review: {e}")
         return
-    clear_previous_review(event)
+    dismiss_previous_reviews(event)
+    threads = get_review_threads(event)
```
UltralyticsAssistant: 💡 **MEDIUM**: This fetch is only reached when the review command runs, but the composite action invokes that command solely for `pull_request`/`pull_request_target` events whose action is `review_requested` (`action.yml:315-328`). An author's reply to an inline review thread therefore never triggers this path, so the bot still cannot actually converse or converge until someone manually re-requests review. Wire the review flow to the review-comment event as well (and gate out the bot's own replies to avoid loops); GitHub exposes inline-comment activity through `pull_request_review_comment`, not `review_requested`. ([docs.github.com](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows?apiVersion=2022-11-28&utm_source=openai))

UltralyticsAssistant (commented):
## 🔍 PR Review

The stale-head safeguard now prevents thread mutations from an outdated review, and the thread query correctly keeps the newest comment tail. However, the action is not invoked when authors reply in review threads, so the central conversational behavior cannot run automatically.

💬 Posted 1 inline comment
🧵 Updated 2 existing review threads

actions/review_pr.py
JaviChulvi: This is the intended design, not a gap. Wiring the review to `pull_request_review_comment` would start a full LLM review run for every individual reply: authors typically leave several thread replies in a burst, and Actions has no debouncing, so a single review pass with 4 replies would trigger 4 complete multi-minute runs with their LLM cost. An author gate does not fully close the loop risk either, since other bots also comment on Ultralytics PRs. The batched turn-based flow is deliberate: reply to any number of threads for free, then one re-request of the bot's review processes them all in a single round — the same one-click affordance maintainers already use. Automatic per-comment execution is out of scope for this PR.

UltralyticsAssistant: The manual re-request workflow intentionally batches replies into one review run, avoiding repeated costly LLM reviews for each individual comment. This is a valid design tradeoff rather than a correctness gap.

actions/review_pr.py
```diff
@@ -361,13 +405,40 @@ def parse_diff_files(diff_text: str) -> tuple[dict, str]:
...
+
+
 def generate_pr_review(
     repository: str,
     diff_text: str,
     pr_title: str,
     pr_description: str,
     event: Action = None,
     head_sha: str | None = None,
+    threads: dict[str, dict] | None = None,
```
UltralyticsAssistant, line 441: 💡 **MEDIUM**: The early-return paths above `threads = threads or {}` do not set `open_threads`. Consequently, a re-request on a PR with no parseable text diff (for example, binary-only or fully skipped generated changes) reaches `post_review_summary()` with unresolved threads but `has_open_threads == False`, so the new review can APPROVE despite the persisted findings. Include the supplied thread count in every early-return result, or initialize and attach `open_threads` before those branches.

actions/review_pr.py
```diff
@@ -687,29 +823,87 @@ def _verified_local_checkout(head_sha: str | None) -> bool:
...
+        if not page_info.get("hasNextPage"):
+            return threads
+        cursor = page_info.get("endCursor")
+
+
+def apply_thread_actions(event: Action, review_data: dict, threads: dict[str, dict]) -> None:
+    """Reply to and resolve existing review threads as decided by the review model."""
+    if not (actions := review_data.get("thread_actions")):
+        return
+    if event.get_pr_head_sha() != review_data["head_sha"]:  # don't mutate threads from a stale review
```
UltralyticsAssistant, line 936: 💡 **MEDIUM**: Checking only the PR head does not make thread actions fresh: an author can reply while the multi-minute model call is running without changing the SHA. The bot can then resolve a thread based on a conversation it did not see, or post a reply after a newer message. Re-fetch/validate each targeted thread's latest comment state immediately before applying actions, and skip the action if it changed.

UltralyticsAssistant (commented):
## 🔍 PR Review

Persistent review threads, action validation, and duplicate-anchor filtering are a solid direction. Two edge cases remain: approval can bypass unresolved threads on early review exits, and thread actions can be applied to a stale conversation.

💬 Posted 2 inline comments
🧵 Updated 1 existing review thread

actions/review_pr.py
JaviChulvi: APPROVE already requires evidence: `post_review_summary` gates on `has_evidence = bool(review_data.get('diff_files'))`, and none of the early-return paths include `diff_files`, so they all fall through to COMMENT regardless of `open_threads`. The missing key is cosmetic, not a behavioral gap — adding it to every early return would be speculative scaffolding.

UltralyticsAssistant: Confirmed. The early-return results omit `diff_files`, so `has_evidence` forces a COMMENT regardless of `open_threads`.

actions/review_pr.py
```diff
@@ -637,10 +750,33 @@ def generate_pr_review(
...
+                continue
+            thread_actions[ref] = {"thread": ref, "action": verb, "message": message}
+        resolved = sum(a["action"] == "resolve" for a in thread_actions.values())
+
         response.update(
             {
                 "comments": filtered_comments,
                 "comments_before_filtering": comments_before_filtering,
+                "thread_actions": list(thread_actions.values()),
+                "open_threads": len(threads) - resolved,
```
UltralyticsAssistant: ⚠️ **HIGH**: `open_threads` is decremented for every model-requested resolve before `apply_thread_actions()` verifies that it can actually perform it. If the freshness check at lines 898-901 skips a resolve because someone replied (or resolved) during generation, this value still reaches zero and `post_review_summary()` can publish an `APPROVE` despite the original thread remaining unresolved. Return the successfully resolved thread IDs/count from `apply_thread_actions()` and derive both `open_threads` and the “Updated” summary count from actions that were actually applied.

UltralyticsAssistant (commented):
## 🔍 PR Review

The persistent-thread flow is well integrated: stale review decisions are dismissed without deleting discussions, thread context is bounded before prompting, and duplicate anchors are filtered. One approval-state bug remains when a requested resolution is skipped during the final freshness check.

💬 Posted 1 inline comment
🧵 Updated 1 existing review thread

UltralyticsAssistant (commented):
## 🔍 PR Review

The PR correctly preserves inline review discussions and adds a conservative open-thread gate for approvals. Two existing issues remain: thread freshness is still checked from a batch snapshot rather than immediately per mutation, and the summary can count skipped thread actions as updates.

actions/review_pr.py
JaviChulvi: The substantive window here was the multi-minute model call, and the batch snapshot closes it: `apply_thread_actions` re-fetches every thread immediately before the loop and skips any whose conversation changed. What remains is the seconds between that fetch and each mutation — and GitHub's API has no compare-and-swap, so a per-mutation re-fetch cannot eliminate that residual window either, only shrink it while adding one query per action. A reply landing in those final seconds is also recoverable: if the bot replied, the human message is still the latest and gets processed next round; if it resolved, one click unresolves and the thread re-enters the flow. This is diminishing-returns hardening against an irreducible race, so we are keeping the batch snapshot.

UltralyticsAssistant: The batch re-fetch now closes the substantive generation-time race before actions are applied. The remaining per-mutation interval is not fully avoidable without compare-and-swap support.

actions/review_pr.py
```diff
@@ -361,13 +405,40 @@ def parse_diff_files(diff_text: str) -> tuple[dict, str]:
...
+    """Format unresolved review threads as a bounded prompt section, returning the refs actually shown."""
+    blocks, shown, total = [], set(), 0
+    for ref, thread in threads.items():
+        anchor = f"{thread['path']}:{thread['line']} ({thread['side']})" if thread.get("line") else thread["path"]
+        outdated = " [outdated: the anchored code changed since]" if thread["outdated"] else ""
+        comments = thread["comments"]
+        if len(comments) > 10:  # root finding plus the latest replies: the newest rebuttal matters most
+            comments = [comments[0], *comments[-9:]]
+        convo = "\n".join(f"  {c['author']}: {c['body'][:1000]}" for c in comments)
+        block = _clip_tool_output(f"[{ref}] {anchor}{outdated}\n{convo}", MAX_THREADS_SECTION_CHARS)
```
UltralyticsAssistant: 💡 **MEDIUM**: A single long thread is clipped to 8,000 characters before it is added to `shown`, despite the intent to preserve whole threads. With the allowed root plus nine replies, this can cut off the newest replies—the very rebuttal the review needs to evaluate—while still allowing the model to resolve or reply to that incomplete thread. Bound the selected conversation without truncating its tail, or omit an oversized thread from `shown` so actions are never based on a partial conversation.

UltralyticsAssistant (commented):
## 🔍 PR Review

The persistent-thread flow is well integrated, including stale-review dismissal, duplicate-anchor filtering, and action freshness checks. One prompt-budget edge case can still make the bot act on a thread without seeing its latest replies.

💬 Posted 1 inline comment
🧵 Updated 1 existing review thread

actions/review_pr.py
UltralyticsAssistant: The stale-resolve path now restores `open_threads` before the summary is posted, so an unresolved thread cannot lead to an APPROVE.

UltralyticsAssistant (commented):
## 🔍 PR Review

The stale-resolve approval gate is now kept conservative, and persistent review threads are no longer deleted. One existing thread-context truncation issue remains.

🧵 Updated 1 existing review thread

actions/review_pr.py
JaviChulvi: Fixed in 3a8fb26: blocks are now built from the root plus the 6 newest replies, so the worst case is ~7.5k chars (7 comments clipped to 1,000 chars each plus the header) and always fits the 8,000-char budget — the per-block `_clip_tool_output` was deleted, so no truncation path remains and `shown` can only contain threads rendered whole.

UltralyticsAssistant: This still truncates every selected comment at `c['body'][:1000]` on line 417. A long newest author rebuttal can therefore lose its substantive conclusion while the model is still allowed to resolve or reply to the thread. The per-thread budget no longer cuts off whole reply entries, but it still presents a partial conversation; preserve the relevant tail or omit oversized conversations rather than acting on clipped replies.

UltralyticsAssistant (commented):
## 🔍 PR Review

The persistent-thread flow is well integrated: stale review decisions are dismissed without deleting discussions, actions are guarded against stale heads and changed conversations, and unresolved threads prevent an approval. One existing thread-budget issue still needs correction.

🧵 Updated 1 existing review thread

actions/review_pr.py
JaviChulvi: Fixed in bb68846: comment bodies over 1,000 chars now carry an explicit `... (truncated)` marker, so the model can see when a reply is partial and decline to act on incomplete text. Blocks are still built whole (root plus the 6 newest replies, ~7.6k worst case within the 8k budget), so no silent truncation path remains. Omitting oversized threads from `shown` would be worse than the disease: the bot could never act on them while they keep blocking approval — a deadlock only a human could break.

UltralyticsAssistant: The marker makes truncation visible, but it does not prevent the model from resolving or replying based on the partial text: the thread rules contain no requirement to decline actions when a comment is truncated, and the action validation does not enforce one. A substantive conclusion after the first 1,000 characters can therefore still be omitted while the thread is acted on, so the original partial-conversation risk remains.

actions/review_pr.py
```diff
@@ -589,6 +706,11 @@ def generate_pr_review(
                 print(f"Filtered out {file_path}:{line_num} (side={side}, available: {available})")
                 continue
 
+            # An existing thread already anchors here; the model must use a thread action, not a duplicate comment
+            if (file_path, side, line_num) in thread_anchors:
```
UltralyticsAssistant: ⚠️ **HIGH**: This treats a coordinate as an existing finding, so it drops every newly discovered issue on a line that already has any unresolved thread—even an unrelated LOW/SUGGESTION thread. The model cannot reliably recover through `thread_actions`: those actions cannot create a distinct finding, and a reply is explicitly discarded below when the root bot comment is still the last comment. In that case, a newly introduced high-severity regression at the same anchor is silently filtered, `open_threads` remains zero for the non-blocking old thread, and the review can approve. Deduplicate against the same finding/thread rather than every `(path, side, line)` anchor, or allow a distinct inline finding at an occupied anchor.

UltralyticsAssistant (commented):
## 🔍 PR Review

The persistent-thread flow, stale-review dismissal, and freshness checks are well structured. One issue remains: the coordinate-only duplicate filter can suppress unrelated new findings and permit an incorrect approval.

💬 Posted 1 inline comment
🧵 Updated 1 existing review thread
🔁 Reply in the threads and re-request my review to continue the conversation

UltralyticsAssistant (commented):
## 🔍 PR Review

The persistent-thread flow is well structured, including stale-head checks and approval gating for unresolved findings. Two existing thread issues still need resolution.

🔁 Reply in the threads and re-request my review to continue the conversation

actions/review_pr.py
UltralyticsAssistant: Dissent noted: visible truncation still permits actions based on partial replies, but this has reached the two-reply debate limit.

UltralyticsAssistant (commented):
## 🔍 PR Review

The PR cleanly preserves prior inline comments and adds freshness checks before mutating threads. One existing thread-deduplication concern still needs attention; the truncation discussion has reached the configured debate limit.

🧵 Updated 1 existing review thread
🔁 Reply in the threads and re-request my review to continue the conversation

actions/review_pr.py
JaviChulvi: The residual after 6a7f055 is self-limiting. At a blocking anchor, suppression is temporary and safe: that same thread already blocks approval, and once it resolves the anchor frees and the finding posts next round. The only permanent case is an equal-or-lower-severity finding behind an advisory nit on the same line — a nit hidden behind a nit, below the materiality bar. Every channel that would let same-severity comments through (semantic matching, model-declared distinctness, rerouting into the thread) reintroduces the re-raising loop this filter exists to prevent. Working as designed.

UltralyticsAssistant: The revised anchor guard now permits a strictly higher-severity finding at an occupied anchor. A same-or-higher existing finding continues to block approval, so the original risk of suppressing a material new finding behind an advisory thread is addressed.

UltralyticsAssistant (approved):
## 🔍 PR Review

The persistent-thread flow is coherent: prior decisions are dismissed without deleting discussions, existing bot threads are fetched and supplied to the reviewer, and thread updates are freshness-checked before the summary is posted. No new actionable issues found in this diff.

🧵 Updated 1 existing review thread

PR #859, opened 2026-08-07
Number PR reviews and carry prior review context forward

glenn-jocher:
## 🛠️ PR Summary

### 🌟 Summary
🤖 Upgrades automated PR reviews to maintain numbered review history and use prior feedback for more consistent, context-aware evaluations.

### 📊 Key Changes
- Bumped the `actions` package version from `0.2.47` to `0.3.0`.
- Added review-history collection, formatting, clipping, and prompt-budget controls.
- Added a `read_pr_conversation` tool so the review agent can inspect earlier reviews, replies, and PR discussion comments.
- Refactored review execution into `run_review()`, which:
  - Captures previous bot reviews.
  - Dismisses superseded review decisions.
  - Removes outdated inline comments.
  - Generates and posts the next numbered review.
- Added persistent findings logs to review summaries so key issues remain available after inline comments are deleted.
- Added GitHub REST API pagination through `Action.paginate()`.
- Updated first-interaction handling and tests to use the new review workflow.

### 🎯 Purpose & Impact
- 🧠 Helps automated reviews avoid repeating, contradicting, or reviving issues that contributors have already addressed.
- 💬 Gives the reviewer access to responses and discussion context, improving decisions on disputed findings.
- 🔢 Makes review iterations easier to track with numbered summaries and commit references.
- 🗂️ Preserves a concise record of prior findings even when superseded inline comments are removed.
- 📈 Improves reliability for PRs with large histories or many GitHub API results through pagination and bounded context handling.
- ✅ Maintains existing stale-PR protection while centralizing review behavior for easier testing and future maintenance.

### DIFF
--- a/actions/first_interaction.py
+++ b/actions/first_interaction.py
@@ -216,14 +216,7 @@ def main(*args, **kwargs):
         # Automatic PR review after first interaction
         if AUTO_PR_REVIEW:
             print("Starting automatic PR review...")
-            try:
-                review_diff, head_sha = event.get_pr_diff_snapshot()
-            except RuntimeError as e:
-                print(f"Skipping stale PR review: {e}")
-                return
-            review_data = review_pr.generate_pr_review(event.repository, review_diff, title, body, event, head_sha)
-            review_pr.post_review_summary(event, review_data)
-            print("PR review completed")
+            review_pr.run_review(event, title, body)
         return
 
     # Handle issues and discussions (NOT PRs)
--- a/actions/review_pr.py
+++ b/actions/review_pr.py
@@ -123,13 +123,19 @@ def _format_review_history(
     """Render prior reviews and the responses to them as review context, oldest dropped first when over budget."""
     sections, reviews = [], history.get("reviews") or []
     if responses := history.get("responses"):
-        sections.append("### Comments from other reviewers\n" + "\n".join(_clip(r, clip) for r in responses))
-    for review in reversed(reviews[-limit:] if limit else reviews):  # newest first, so it survives the budget
+        section = "### Comments from other reviewers\n" + "\n".join(_clip(r, clip) for r in responses)
+        sections.append(_clip(section, budget // 2 if budget else None))  # leave half the budget for the reviews
+
+    shown = reviews[-limit:] if limit else reviews
+    omitted = len(reviews) - len(shown)
+    for review in reversed(shown):  # newest first, so it survives the budget
         section = f"### Review {review['number']} (commit {review['commit']})\n{_clip(review['body'], clip)}"
         if budget and sum(len(s) + 2 for s in sections) + len(section) > budget:
-            sections.append(f"### {review['number']} older review(s) omitted for length")
+            omitted = review["number"]
             break
         sections.append(section)
+    if omitted:
+        sections.append(f"### {omitted} older review(s) omitted for length")
     return "\n\n".join(reversed(sections))
 
 
--- a/actions/review_pr.py
+++ b/actions/review_pr.py
@@ -122,27 +122,36 @@ def _build_review_history(reviews: list[dict], comments: list[dict], bot_usernam
     return {"reviews": prior, "replies": replies, "others": others}
 
 
+def _fit(entries: list[str], budget: int | None, separator: int = 1) -> tuple[list[str], int]:
+    """Keep the newest entries that fit the budget, oldest first, with the count of those dropped."""
+    kept, used = [], 0
+    for entry in reversed(entries):
+        if budget and used + len(entry) + separator > budget:
+            break
+        kept.append(entry)
+        used += len(entry) + separator
+    return kept[::-1], len(entries) - len(kept)
+
+
 def _format_review_history(
     history: dict, limit: int | None = None, clip: int | None = None, budget: int | None = None
 ) -> str:
     """Render prior reviews and the responses to them as review context, oldest dropped first when over budget."""
-    sections, reviews = [], history.get("reviews") or []
-    for title, key in (("Replies to your findings", "replies"), ("Other review comments on this PR", "others")):
+    sections = []
+    for title, key in (("Other review comments on this PR", "others"), ("Replies to your findings", "replies")):
         if entries := history.get(key):
-            section = f"### {title}\n" + "\n".join(_clip(e, clip) for e in entries)
-            sections.append(_clip(section, budget // 4 if budget else None))  # leave half the budget for the reviews
+            kept, dropped = _fit([_clip(e, clip) for e in entries], budget // 4 if budget else None)
+            note = f" ({dropped} older comment(s) omitted for length)" if dropped else ""
+            sections.append(f"### {title}{note}\n" + "\n".join(kept))
 
+    reviews = history.get("reviews") or []
     shown = reviews[-limit:] if limit else reviews
-    omitted = len(reviews) - len(shown)
-    for review in reversed(shown):  # newest first, so it survives the budget
-        section = f"### Review {review['number']} (commit {review['commit']})\n{_clip(review['body'], clip)}"
-        if budget and sum(len(s) + 2 for s in sections) + len(section) > budget:
-            omitted = review["number"]
-            break
-        sections.append(section)
-    if omitted:
-        sections.append(f"### {omitted} older review(s) omitted for length")
-    return "\n\n".join(reversed(sections))
+    rendered = [f"### Review {r['number']} (commit {r['commit']})\n{_clip(r['body'], clip)}" for r in shown]
+    remaining = budget - sum(len(s) + 2 for s in sections) if budget else None
+    kept, dropped = _fit(rendered, remaining, separator=2)
+    if omitted := dropped + len(reviews) - len(shown):
+        kept.insert(0, f"### {omitted} older review(s) omitted for length")
+    return "\n\n".join(kept + sections)
 
 
 def _fetch_head_file(event: Action, sha: str, path: str) -> str | None:
--- a/actions/review_pr.py
+++ b/actions/review_pr.py
@@ -242,13 +242,9 @@ def build_review_agent_tools(
             return f"{path} has no diff lines in requested range {start}-{end}."
         return _clip_tool_output(f"{path} diff:{start}-{end}\n" + "\n".join(lines[start - 1 : end]))
 
-    def read_pr_conversation(review_number=None) -> str:
-        """Read prior reviews of this PR in full, plus the human discussion on it."""
-        selected = {
-            "reviews": [r for r in review_history.get("reviews") or [] if review_number in (None, r["number"])],
-            "responses": review_history.get("responses"),
-        }
-        parts = [_format_review_history(selected) or f"No prior review {review_number}."]
+    def read_pr_conversation() -> str:
+        """Read every prior review of this PR in full, plus the discussion on it."""
+        parts = [_format_review_history(review_history)]
         if event and (pr_number := (event.pr or {}).get("number")) and not discussion:
             url = f"{GITHUB_API_URL}/repos/{event.repository}/issues/{pr_number}/comments"
             response = event.get(url, params={"per_page": 100})
@@ -306,18 +302,11 @@ def build_review_agent_tools(
             "type": "function",
             "name": "read_pr_conversation",
             "description": (
-                "Read your earlier reviews of this PR in full - their summaries, inline findings, and the replies "
+                "Read every earlier review of this PR in full - their summaries, inline findings, and the replies "
                 "to them - plus the PR discussion comments. Use this before repeating, contradicting, or "
                 "reversing an earlier finding, and whenever the PRIOR REVIEWS excerpt in the prompt is truncated."
             ),
-            "parameters": {
-                "type": "object",
-                "properties": {
-                    "review_number": {"type": ["integer", "null"], "description": "One prior review number, or null."},
-                },
-                "required": ["review_number"],
-                "additionalProperties": False,
-            },
+            "parameters": {"type": "object", "properties": {}, "required": [], "additionalProperties": False},
             "strict": True,
         },
         {
--- a/actions/review_pr.py
+++ b/actions/review_pr.py
@@ -95,6 +95,7 @@ def _clip(text: str, limit: int | None) -> str:
 def _build_review_history(reviews: list[dict], comments: list[dict], bot_username: str | None) -> dict:
     """Capture prior bot reviews and the responses to them; each review body carries its own findings log."""
     parents = {comment.get("id"): comment for comment in comments}
+    owned_ids = {review.get("id") for review in reviews}
     prior = []
     for number, review in enumerate(reviews, 1):
         body = re.sub(rf"{re.escape(REVIEW_MARKER)} *\d*", "", review.get("body") or "").replace(ACTIONS_CREDIT, "")
@@ -106,15 +107,19 @@ def _build_review_history(reviews: list[dict], comments: list[dict], bot_usernam
             }
         )
 
-    responses = []
+    replies, others = [], []
     for comment in comments:
         if (login := comment.get("user", {}).get("login")) == bot_username:
             continue
         parent = parents.get(comment.get("in_reply_to_id")) or {}
-        replying_to = f' (replying to "{_clip((parent.get("body") or "").strip(), 120)}")' if parent else ""
         line = comment.get("line") or comment.get("original_line") or 0
-        responses.append(f"- {comment.get('path')}:{line} @{login}{replying_to}: {(comment.get('body') or '').strip()}")
-    return {"reviews": prior, "responses": responses}
+        entry = f"- {comment.get('path')}:{line} @{login}"
+        body = (comment.get("body") or "").strip()
+        if parent.get("pull_request_review_id") in owned_ids:  # a reply to one of our own findings
+            replies.append(f'{entry} on "{_clip((parent.get("body") or "").strip(), 120)}": {body}')
+        else:
+            others.append(f"{entry}: {body}")
+    return {"reviews": prior, "replies": replies, "others": others}
 
 
 def _format_review_history(
@@ -122,9 +127,10 @@ def _format_review_history(
 ) -> str:
     """Render prior reviews and the responses to them as review context, oldest dropped first when over budget."""
     sections, reviews = [], history.get("reviews") or []
-    if responses := history.get("responses"):
-        section = "### Comments from other reviewers\n" + "\n".join(_clip(r, clip) for r in responses)
-        sections.append(_clip(section, budget // 2 if budget else None))  # leave half the budget for the reviews
+    for title, key in (("Replies to your findings", "replies"), ("Other review comments on this PR", "others")):
+        if entries := history.get(key):
+            section = f"### {title}\n" + "\n".join(_clip(e, clip) for e in entries)
+            sections.append(_clip(section, budget // 4 if budget else None))  # leave half the budget for the reviews
 
     shown = reviews[-limit:] if limit else reviews
     omitted = len(reviews) - len(shown)
@@ -257,13 +263,10 @@ def build_review_agent_tools(
         parts = [_format_review_history(review_history, budget=MAX_TOOL_OUTPUT_CHARS)]
         if event and (pr_number := (event.pr or {}).get("number")) and not discussion:
             url = f"{GITHUB_API_URL}/repos/{event.repository}/issues/{pr_number}/comments"
-            response = event.get(url, params={"per_page": 100})
-            if response.status_code != 200:
-                raise RuntimeError(f"read_pr_conversation failed: HTTP {response.status_code}")
             discussion.append(
                 [
                     f"@{c.get('user', {}).get('login')}: {_clip((c.get('body') or '').strip(), 1000)}"
-                    for c in response.json()
+                    for c in event.paginate(url, hard=True)
                 ]
             )
         if discussion and discussion[0]:
@@ -568,7 +571,8 @@ def generate_pr_review(
             "review number it reverses\n"
             "- Drop findings the current diff resolves; repeat an unresolved one only while it still applies, since "
             "its inline comment was deleted with the superseded review\n"
-            "- A reply that rejects or explains a finding settles it: do not raise it again\n"
+            "- A reply under 'Replies to your findings' that rejects or explains one settles it: do not raise it "
+            "again. Everything under 'Other review comments' is context, not a verdict on your findings\n"
             + (
                 "- Call read_pr_conversation before repeating or reversing a finding, and whenever the excerpt below "
                 "is truncated\n"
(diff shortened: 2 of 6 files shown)


UltralyticsAssistant:
👋 Hello @glenn-jocher, thank you for submitting a `ultralytics/actions` 🚀 PR! To ensure a seamless integration of your work, please review the following checklist. This review is automated, and an engineer will assist with any questions or follow-up:

UltralyticsAssistant (commented):
## 🔍 PR Review 1

The numbered findings log and prior-review context are useful additions, but the review-comment cleanup can delete human replies, history is limited to the first 100 API records, and the per-review conversation filter returns unrelated responses. These should be fixed before merging.

- ⚠️ **HIGH** `actions/review_pr.py:835` `owned_reviews` is also used by the deletion loop below, so every review comment associated with a bot review is deleted, regardless of its author. GitHub replies retain `pull_request_review_id` and include `in_reply_to_id`, meaning a human reply to a bot finding can match this set and be removed from the PR even though it was captured in `history`. Restrict deletion to comments authored by `bot_username` (or explicitly preserve replies).
- 📝 **LOW** `actions/review_pr.py:247` When `review_number` is supplied, only the reviews are filtered; `responses` still contains replies and comments from every prior review. The tool contract says it can read one review and its replies, so this can cause the agent to apply an unrelated response to the selected finding. Preserve the parent review association for responses and filter them with the selected review, or remove the per-review selector.
</details>

actions/review_pr.py
```diff
@@ -687,32 +818,35 @@ def _verified_local_checkout(head_sha: str | None) -> bool:
...
         if review.get("user", {}).get("login") == bot_username and REVIEW_MARKER in (review.get("body") or "")
-    }
-    for review in reviews:
-        if review.get("id") in owned_reviews and review.get("state") in ("APPROVED", "CHANGES_REQUESTED"):
+    ]
+    comments_base = f"{GITHUB_API_URL}/repos/{event.repository}/pulls/{pr_number}/comments"
+    comments = event.get(comments_base, params={"per_page": 100}, hard=True).json()
+    history = _build_review_history(owned, comments, bot_username)  # capture responses before deleting the comments
+
+    owned_reviews = {review["id"] for review in owned}
```
glenn-jocher, line 848: Checked this against real GitHub data before touching the deletion loop: replies do not inherit the parent comment’s `pull_request_review_id`. Across the 100 most recent review comments in ultralytics/ultralytics and ultralytics/actions, every reply carried its own review id, distinct from its parent’s (e.g. reply 3732985415 review=4879390280 -> parent review=4879324596). GitHub creates a separate single-comment review for each reply, so a human reply cannot match `owned_reviews`, which only holds bot reviews whose body carries the review marker. No change made — an author guard here would be a speculative check for a state that cannot occur.

actions/review_pr.py
```diff
@@ -193,6 +242,28 @@ def read_diff(path: str, start_line=None, end_line=None) -> str:
             return f"{path} has no diff lines in requested range {start}-{end}."
         return _clip_tool_output(f"{path} diff:{start}-{end}\n" + "\n".join(lines[start - 1 : end]))
 
+    def read_pr_conversation(review_number=None) -> str:
+        """Read prior reviews of this PR in full, plus the human discussion on it."""
+        selected = {
```
glenn-jocher: Correct, the selector filtered reviews but not responses. Fixed in 3914c10 by removing the `review_number` parameter: the tool now always returns the full history plus the PR discussion, which fits the 20k tool-output cap, dropping the mismatched filter instead of adding a second one.

UltralyticsAssistant (commented):
## 🔍 PR Review 2

Since the previous review, the per-review selector issue was removed, and the concern about deleting human replies was investigated and is not supported by the repository's GitHub data. The numbered findings log and continuity context are useful, but pagination remains unresolved and this revision introduces two context-budget problems: response history can exceed the configured limit, and older reviews can be silently omitted without telling the agent.

- 💡 **MEDIUM** `actions/review_pr.py:126` `MAX_HISTORY_CHARS` is not enforced for responses: this joins every response, clipped individually to 8,000 characters, before the budget check is applied only to review sections below. Three sufficiently long responses can already exceed the 20,000-character history budget, while `diff_budget` bottoms out at 1,000 and the final prompt still exceeds `MAX_PROMPT_CHARS`, potentially causing request failures. Apply the aggregate budget while adding responses as well as reviews.
- 💡 **MEDIUM** `actions/review_pr.py:127` When `limit=MAX_HISTORY_REVIEWS` is used, `reviews[-limit:]` silently drops older reviews without adding an omission marker. The continuity instructions tell the agent to call `read_pr_conversation` when the excerpt is truncated, but after more than five reviews the prompt gives no indication that history is missing, so findings from older reviews can be repeated. Preserve an explicit omitted count or otherwise signal that the excerpt is incomplete.
</details>

actions/review_pr.py
```diff
@@ -83,6 +87,52 @@ def search_repo(query: str, path_glob=None) -> str:
...
+    return {"reviews": prior, "responses": responses}
+
+
+def _format_review_history(
+    history: dict, limit: int | None = None, clip: int | None = None, budget: int | None = None
+) -> str:
+    """Render prior reviews and the responses to them as review context, oldest dropped first when over budget."""
+    sections, reviews = [], history.get("reviews") or []
+    if responses := history.get("responses"):
+        sections.append("### Comments from other reviewers\n" + "\n".join(_clip(r, clip) for r in responses))
```
glenn-jocher: Fixed in 89884cc: the responses section is now clipped to half the budget before the review sections are added, and the running budget check already counts it, so the whole excerpt stays within `MAX_HISTORY_CHARS`. Verified with five 9k responses plus five 3k reviews: total 19,130 chars, responses 9,968.

actions/review_pr.py
```diff
@@ -83,6 +87,52 @@ def search_repo(query: str, path_glob=None) -> str:
...
+    for review in reversed(reviews[-limit:] if limit else reviews):  # newest first, so it survives the budget
```
glenn-jocher: Fixed in 89884cc: the omitted count now covers reviews dropped by `MAX_HISTORY_REVIEWS` as well as by the budget, so `### N older review(s) omitted for length` appears in both cases. Verified with 8 reviews at limit=5 and an ample budget: the marker reports 3.

UltralyticsAssistant (commented):
## 🔍 PR Review 3

Since the previous review, the response-budget and omitted-review markers were addressed, and the rejected human-reply deletion concern remains resolved. Two issues remain: response history is not scoped to replies to this bot's reviews, and the new history/cleanup reads only the first 100 records from each GitHub collection, so continuity and cleanup fail on sufficiently active PRs.

<details><summary>💬 Posted 1 inline comment</summary>

- 💡 **MEDIUM** `actions/review_pr.py:113` Every non-bot review comment is added to `responses`; `parent` only changes the display text and does not restrict inclusion. This means top-level human comments and comments on other reviewers' or automations' reviews are presented as responses to this bot's findings, so the continuity rule that a reply settles a finding can suppress or distort future findings. Include only comments whose `in_reply_to_id` resolves to a comment belonging to one of `owned` reviews, and keep unrelated review comme…
</details>

actions/review_pr.py
```diff
@@ -83,6 +87,58 @@ def search_repo(query: str, path_glob=None) -> str:
...
+                "commit": (review.get("commit_id") or "")[:7],
+                "body": body.split("<details><summary>📋 Skipped")[0].strip(),
+            }
+        )
+
+    responses = []
+    for comment in comments:
+        if (login := comment.get("user", {}).get("login")) == bot_username:
+            continue
+        parent = parents.get(comment.get("in_reply_to_id")) or {}
```
glenn-jocher, line 114: Fixed in aacce56: history now splits into `replies` (parent comment belongs to one of the bot's own reviews) and `others`, rendered as separate sections, and the continuity rule scopes "settles it" to the replies section while labelling the rest as context, not a verdict. The 100-record cap you flagged in the summary is fixed in the same commit: `Action.paginate` now walks every page for the reviews, review comments, and PR discussion reads.

UltralyticsAssistant (commented):
## 🔍 PR Review 4

Prior review concerns about response scoping and collection pagination are addressed, and the numbered findings log is useful. Two issues remain: the new conversation tool cannot retrieve history omitted by its own output budget, and arbitrary non-bot replies can be treated as authoritative resolutions and suppress future findings.

- 💡 **MEDIUM** `actions/review_pr.py:263` This fallback does not actually make the full history available once it exceeds 20,000 characters: `_format_review_history(..., budget=MAX_TOOL_OUTPUT_CHARS)` still drops older review sections, and `_clip_tool_output` can truncate the result again. Because the tool has no paging or review-range arguments, an agent that sees an omitted-history marker cannot retrieve those omitted findings, so older unresolved issues may still be repeated or lost. Return the history in requestable chunks or add pa…
- 💡 **MEDIUM** `actions/review_pr.py:574` The replies in this section come from every non-bot review commenter, including the PR author and other automations; no author trust or maintainer permission is checked before this instruction treats a reply as settling a finding. A contributor can therefore reply that an unresolved defect is intentional and cause the next review to suppress the finding. Keep untrusted author replies as context, or only let trusted reviewer/maintainer responses settle prior findings.
</details>

actions/review_pr.py
```diff
@@ -193,6 +258,21 @@ def read_diff(path: str, start_line=None, end_line=None) -> str:
             return f"{path} has no diff lines in requested range {start}-{end}."
         return _clip_tool_output(f"{path} diff:{start}-{end}\n" + "\n".join(lines[start - 1 : end]))
 
+    def read_pr_conversation() -> str:
+        """Read every prior review of this PR in full, plus the discussion on it."""
+        parts = [_format_review_history(review_history, budget=MAX_TOOL_OUTPUT_CHARS)]
```
glenn-jocher, line 274: Declining this one. The tool already returns the newest reviews first and states `### N older review(s) omitted for length` when it drops any, so nothing is silently lost. Reaching that cap needs more than 20k characters of review bodies on a single PR — roughly six full reviews — and a PR that long has bigger problems than history paging. A chunking or range API is scaffolding for a case that does not occur, and this repo's principles put deletion and reuse ahead of new surface. If it ever does bite, the fix is a larger budget constant, not a new tool contract.

actions/review_pr.py
```diff
@@ -452,6 +559,29 @@ def generate_pr_review(
...
+            f"CONTINUITY - you already reviewed this PR {len(prior_reviews)} time(s), this is review "
+            f"{len(prior_reviews) + 1}:\n"
+            "- PRIOR REVIEWS below carries your earlier summaries and findings plus the replies to them; the diff "
+            "shows the current state\n"
+            "- Never contradict yourself without cause: a change made to satisfy an earlier finding is not a new "
+            "problem, and an earlier finding is reversed only with concrete new evidence, stated as such with the "
+            "review number it reverses\n"
+            "- Drop findings the current diff resolves; repeat an unresolved one only while it still applies, since "
+            "its inline comment was deleted with the superseded review\n"
+            "- A reply under 'Replies to your findings' that rejects or explains one settles it: do not raise it "
```
glenn-jocher, line 585: Partly taken. A permission or trust gate would mean extra API calls and a policy layer inside the reviewer, which is scaffolding this repo avoids — and the review is advisory anyway, with the whole thread visible to the human who merges. What was genuinely missing is who is speaking, so the reply lines now carry the GitHub author association: `- path:line @user (MEMBER) on "...": ...`. The model sees OWNER/MEMBER/COLLABORATOR/CONTRIBUTOR/NONE and can weigh a drive-by "this is intentional" against a maintainer's, without a hardcoded trust rule.

UltralyticsAssistant (commented):
## 🔍 PR Review 5

Prior concerns about pagination, response scoping, aggregate history budgeting, and omitted-review markers are addressed. One continuity issue remains: oversized reply sections discard later responses without clearly indicating the omitted context, which can cause an already-resolved finding to be repeated.

<details><summary>💬 Posted 1 inline comment</summary>

- 💡 **MEDIUM** `actions/review_pr.py:133` When either response list exceeds `budget // 4`, this clips the already-joined string from the front. Later entries—including a response that may settle a finding—are silently dropped, and no omission marker tells the model what was omitted. Apply the budget while retaining the newest responses (or otherwise preserve all relevant entries) and emit an explicit truncation marker so continuity decisions are not based on incomplete replies.
</details>

actions/review_pr.py
```diff
@@ -83,6 +87,64 @@ def search_repo(query: str, path_glob=None) -> str:
...
+
+def _format_review_history(
+    history: dict, limit: int | None = None, clip: int | None = None, budget: int | None = None
+) -> str:
+    """Render prior reviews and the responses to them as review context, oldest dropped first when over budget."""
+    sections, reviews = [], history.get("reviews") or []
+    for title, key in (("Replies to your findings", "replies"), ("Other review comments on this PR", "others")):
+        if entries := history.get(key):
+            section = f"### {title}\n" + "\n".join(_clip(e, clip) for e in entries)
+            sections.append(_clip(section, budget // 4 if budget else None))  # leave half the budget for the reviews
```
glenn-jocher: Correct — the responses path had the same front-clipping bug the reviews path already fixed. Both now go through one `_fit` helper that keeps the newest entries within the budget and returns the dropped count, so the sections carry `(N older comment(s) omitted for length)` and `### N older review(s) omitted for length`. Verified with five 3k replies at a 20k budget: the newest reply survives, the oldest four are dropped and counted.

UltralyticsAssistant (commented):
## 🔍 PR Review 6

Since the previous review, the response-budget clipping issue now retains newest entries and reports omissions, and the earlier pagination, scoping, and cleanup concerns are addressed. Two edge cases remain: oversized newest replies can cause the entire response section to be dropped, and the pagination helper falsely fails at exactly 10,000 items.

- 💡 **MEDIUM** `actions/review_pr.py:143` Each entry is clipped to `MAX_HISTORY_ITEM_CHARS` (8,000), but each response section is limited to `budget // 4` (5,000 with the current 20,000-character budget). If the newest reply exceeds that section budget, `_fit` breaks immediately and retains no replies, while the omission count incorrectly labels the newest reply as older. A long reply that settles a finding can therefore be unavailable to the model. Clip the newest entry to the section budget, or skip oversized entries and continue fitt…
- 📝 **LOW** `actions/utils/github_utils.py:175` After page 100 contains 100 items, this raises without checking whether page 101 is empty. A collection with exactly 10,000 items therefore fails even though pagination is complete, causing `clear_previous_review` to abort before publishing the new review. Follow the response's next-page link or probe one additional page before raising.
</details>

actions/review_pr.py
```diff
@@ -83,6 +87,73 @@ def search_repo(query: str, path_glob=None) -> str:
...
+
+
+def _format_review_history(
+    history: dict, limit: int | None = None, clip: int | None = None, budget: int | None = None
+) -> str:
+    """Render prior reviews and the responses to them as review context, oldest dropped first when over budget."""
+    sections = []
+    for title, key in (("Other review comments on this PR", "others"), ("Replies to your findings", "replies")):
+        if entries := history.get(key):
+            kept, dropped = _fit([_clip(e, clip) for e in entries], budget // 4 if budget else None)
```
glenn-jocher: Both real, fixed in the latest commit. The newest response is now clipped to the section share rather than dropped, which also required two definition fixes the repro exposed: `_clip` overshot its limit by the ellipsis character, and `_fit` charged a separator for the first entry. Verified: a single 9k reply survives clipped to 5,000 chars with no omission marker, and five 3k replies keep the newest with `(4 older comment(s) omitted for length)`, total 18,286 of a 20,000 budget.

actions/utils/github_utils.py
```diff
@@ -164,6 +164,16 @@ def get(self, url, **kwargs):
...
 
+    def paginate(self, url, params=None, **kwargs) -> list:
+        """Fetch every page of a GitHub REST collection."""
+        items = []
+        for page in range(1, 101):
+            page_items = self.get(url, params={**(params or {}), "per_page": 100, "page": page}, **kwargs).json()
+            items.extend(page_items)
+            if len(page_items) < 100:
+                return items
+        raise RuntimeError(f"GitHub collection exceeded 10,000 items: {url}")
```
glenn-jocher, line 175: Fixed: the loop now runs one page past the cap, so a collection of exactly 10,000 items sees an empty page 101 and returns normally. Verified with 100 full pages plus an empty one: 10,000 items, no error.

UltralyticsAssistant (commented):
## 🔍 PR Review 7

The prior response-budget, response-scoping, pagination coverage, and exact-10,000-item fixes remain addressed. Two issues remain: the first upgraded review loses inline findings from pre-existing bot reviews, and pagination silently returns incomplete collections just above the 10,000-item cap. The numbering and current findings-log design otherwise look sound.

- 💡 **MEDIUM** `actions/review_pr.py:112` On the first run after upgrading, older bot reviews created before the findings log was added still have their inline findings only in `comments`. This branch skips those bot-authored comments, so they are absent from `history`; the deletion loop below then removes them, and future reviews cannot carry those findings forward. Include each owned bot comment in its review history before deletion, while keeping human replies in the separate replies section.
- 📝 **LOW** `actions/utils/github_utils.py:173` The cap is not enforced for collections with 10,001–10,099 items: after 100 full pages, a short page 101 causes this to return the partial collection instead of raising. `clear_previous_review` can therefore silently miss older reviews or comments. Check the accumulated count before returning, and raise whenever it exceeds 10,000.
</details>

actions/review_pr.py
```diff
@@ -83,6 +87,75 @@ def search_repo(query: str, path_glob=None) -> str:
...
+            {
+                "number": number,
+                "commit": (review.get("commit_id") or "")[:7],
+                "body": body.split("<details><summary>📋 Skipped")[0].strip(),
+            }
+        )
+
+    replies, others = [], []
+    for comment in comments:
+        if (login := comment.get("user", {}).get("login")) == bot_username:
```
glenn-jocher, line 112: Declining this one. It is a one-time migration case: reviews posted before this PR have no findings log, so their findings are thinner for exactly one round, after which every review body carries its own log. Pulling live bot comments into history as well would duplicate every finding of the newest review (body log plus inline copy) and need dedup to avoid it — a compatibility shim plus extra machinery for a transitional round, which this repo's principles rule out. The prior summaries still carry forward, so nothing is lost outright.

actions/utils/github_utils.py
```diff
@@ -164,6 +164,16 @@ def get(self, url, **kwargs):
         """Performs GET request with error handling."""
         return self._request("get", url, **kwargs)
 
+    def paginate(self, url, params=None, **kwargs) -> list:
+        """Fetch every page of a GitHub REST collection."""
+        items = []
+        for page in range(1, 102):  # one page past the cap, so an exactly-10,000-item collection still completes
+            page_items = self.get(url, params={**(params or {}), "per_page": 100, "page": page}, **kwargs).json()
+            items.extend(page_items)
+            if len(page_items) < 100:
```
glenn-jocher, line 173: Refuting this one on the page arithmetic. Page 101 covers items 10,001-10,100, so a short page 101 means the collection ended inside that range and everything was fetched — 10,001-10,099 items return complete, not partial. The raise fires only when page 101 is also full, i.e. 10,100+, which is the case where data really would be missing. Verified against the code: 10,000 -> 10,000 returned, 10,050 -> 10,050, 10,099 -> 10,099, 10,100 -> RuntimeError, 10,500 -> RuntimeError.

UltralyticsAssistant (commented):
## 🔍 PR Review 8

Earlier pagination, response scoping, and history-budget concerns remain addressed. The numbered findings log and conversation tool are otherwise well integrated, but the automatic first-interaction review still bypasses the new numbering and cleanup flow, allowing duplicate Review 1 entries on reruns.

<details><summary>💬 Posted 1 inline comment</summary>

- 💡 **MEDIUM** `actions/review_pr.py:863` `post_review_summary` is also called directly by `actions/first_interaction.py`, but that path neither calls `clear_previous_review` nor passes a computed review number/history. It therefore always uses the new default `review_number=1`; if the opened-event job is rerun or invoked when a bot review already exists, it publishes another `## 🔍 PR Review 1` and leaves the prior inline comments in place, breaking numbering and continuity. Route that path through the same history/cleanup flow or deri…
</details>

actions/review_pr.py
```diff
@@ -687,32 +832,35 @@ def _verified_local_checkout(head_sha: str | None) -> bool:
...
-    comments_base = f"{GITHUB_API_URL}/repos/{event.repository}/pulls/{pr_number}/comments"
-    comments = event.get(comments_base, params={"per_page": 100}, hard=True).json()
     for comment in comments:
         if comment.get("pull_request_review_id") in owned_reviews:
             event.delete(f"{GITHUB_API_URL}/repos/{event.repository}/pulls/comments/{comment['id']}", hard=True)
+    return history
 
 
-def post_review_summary(event: Action, review_data: dict) -> None:
+def post_review_summary(event: Action, review_data: dict, review_number: int = 1) -> None:
```
glenn-jocher, line 862: Right, and it was a duplication problem rather than a missing argument: both entry points repeated the generate-then-post sequence, so only one of them gained cleanup and numbering. Fixed by extracting `review_pr.run_review(event, title, body)` — diff snapshot, `clear_previous_review`, generate, publish with `review_number=len(history["reviews"]) + 1` — and calling it from both `review_pr.main` and `first_interaction`. The duplicated block in first_interaction is deleted, so a rerun of the opened job now supersedes and renumbers like any other round.

UltralyticsAssistant (dismissed):
## 🔍 PR Review 9

Since the previous review, the first-interaction path has been routed through `run_review`, so review cleanup, history capture, and numbering are now shared with the regular review entry point. The current diff does not introduce any additional evidence-backed issues.

UltralyticsAssistant (commented):
## 🔍 PR Review 10

Earlier continuity, numbering, pagination, and first-interaction routing findings are addressed. The current diff introduces one avoidable performance regression: automatic first-interaction reviews fetch the PR diff twice.

<details><summary>💬 Posted 1 inline comment</summary>

- 📝 **LOW** `actions/first_interaction.py:219` `event.get_pr_diff()` was already fetched at line 198 for the unified PR-open response, but `run_review()` unconditionally calls `get_pr_diff_snapshot()`, which performs another diff request plus two PR-head requests. Every automatically reviewed opened PR now downloads the diff twice and adds avoidable API latency and rate-limit usage. Reuse the existing diff/head snapshot in the shared runner, or restructure the first-interaction path to fetch it only once.
</details>

actions/first_interaction.py
```diff
@@ -216,14 +216,7 @@ def main(*args, **kwargs):
...
             print("Starting automatic PR review...")
-            try:
-                review_diff, head_sha = event.get_pr_diff_snapshot()
-            except RuntimeError as e:
-                print(f"Skipping stale PR review: {e}")
-                return
-            review_data = review_pr.generate_pr_review(event.repository, review_diff, title, body, event, head_sha)
-            review_pr.post_review_summary(event, review_data)
-            print("PR review completed")
+            review_pr.run_review(event, title, body)
```
glenn-jocher, line 219: Refuting this one on both points. It is not introduced here: on main at the branch point (520eff3) the same block already called `event.get_pr_diff()` at line 198 and then `event.get_pr_diff_snapshot()` in the review branch — this PR only moved that identical call into `run_review`, so the request count is unchanged.

The second fetch is also deliberate rather than redundant. `get_pr_diff_snapshot` reads the head SHA, refetches the diff with `refresh=True`, rereads the head, and retries if it moved, which is what lets `post_review_summary` publish against a verified `commit_id` and abort on "PR head changed during review generation". The cached diff from line 198 carries no verified head, so reusing it would trade the staleness guarantee for one saved request on PR-open only.

UltralyticsAssistant (approved):
## 🔍 PR Review 11

Since review 10, the first-interaction path remains correctly routed through the shared review flow, and the prior diff-fetch concern has already been explained as deliberate and unchanged. Numbering, history capture, pagination, findings logging, and conversation-tool integration are consistent in the current diff. I found no new evidence-backed issues.

UltralyticsAssistant:
🎉 Thank you, @glenn-jocher, and congratulations on this excellent merge!

This PR brings together review history, conversation context, pagination, and persistent findings to make automated reviews more consistent, traceable, and useful over time. Your work helps contributors receive smarter feedback while preserving the context that makes each iteration better—an important step forward for reliable, context-aware PR reviews!

PR #876, opened 2026-08-15
Resolve Dart packages before formatting

glenn-jocher:
`dart format` reads the package language version from `.dart_tool/package_config.json`; on a fresh checkout without `pub get` it silently falls back to the pre-3.7 short style, so the bot reformatted 13 files in ultralytics/yolo-flutter-app#585 into the wrong style. Install Flutter from the official flutter/flutter repository (`git clone --depth 1 --branch stable`, as documented by Flutter; no third-party action) — it bundles Dart and is required to resolve Flutter packages — and run `flutter pub get` in every package before formatting. Verified locally with the cloned stable SDK (Dart 3.13.0) against yolo-flutter-app: without package config → short style, with → correct tall style.

## 🛠️ PR Summary

### 🌟 Summary
The Dart-formatting workflow now installs Flutter, resolves every package before formatting, and cleans up generated package metadata to ensure formatting uses the correct Dart language version.

### 📊 Key Changes
- Replaced `dart-lang/setup-dart` with a shallow clone of the official Flutter stable repository and added its bundled Dart SDK to `GITHUB_PATH`.
- Runs `flutter pub get` in every non-hidden directory containing `pubspec.yaml` before invoking `dart format`.
- Skips formatting when package resolution fails, while preserving the action’s existing `continue-on-error` behavior.
- Removes generated files and restores tracked `pubspec.lock` files after formatting.

### 🎯 Purpose & Impact
- Pull requests using Dart formatting now resolve Flutter and Dart package metadata first, preventing formatting with the fallback pre-3.7 style.
- The workflow applies this behavior to `opened` and `synchronize` pull request events when the Dart input is enabled.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -253,12 +253,15 @@ runs:
       if: github.event_name == 'pull_request' && inputs.dart == 'true' && (github.event.action == 'opened' || github.event.action == 'synchronize')
       run: |
         echo "::group::Run Dart Formatter"
-        # Formatting never creates files: always drop generated pub/Flutter metadata and restore tracked lockfiles
-        trap 'git clean -fdqx; git ls-files -z "*pubspec.lock" | xargs -0 -r git checkout --; echo "::endgroup::"' EXIT
+        trap 'echo "::endgroup::"' EXIT
+        ok=true
         while IFS= read -r -d '' pubspec; do
-          (cd "$(dirname "$pubspec")" && flutter pub get)
+          (cd "$(dirname "$pubspec")" && flutter pub get) || ok=false
         done < <(find . -name pubspec.yaml -not -path '*/.*' -print0)
-        dart format .
+        $ok && dart format . || ok=false # never format with an unresolved language version
+        git clean -fdqx # formatting never creates files, so drop generated pub and Flutter metadata
+        git ls-files -z '*pubspec.lock' | xargs -0 -r git checkout --
+        $ok
       shell: bash
       continue-on-error: true # Important so that commit step can pick up changes
 
--- a/action.yml
+++ b/action.yml
@@ -254,6 +254,7 @@ runs:
       run: |
         echo "::group::Run Dart Formatter"
         trap 'echo "::endgroup::"' EXIT
+        echo ".dart_tool/" >> .git/info/exclude # keep resolution artifacts out of the auto-format commit
         find . -name pubspec.yaml -not -path '*/.*' -execdir flutter pub get \;
         dart format .
       shell: bash
--- a/action.yml
+++ b/action.yml
@@ -254,9 +254,12 @@ runs:
       run: |
         echo "::group::Run Dart Formatter"
         trap 'echo "::endgroup::"' EXIT
-        echo ".dart_tool/" >> .git/info/exclude # keep resolution artifacts out of the auto-format commit
-        find . -name pubspec.yaml -not -path '*/.*' -execdir flutter pub get \;
+        for pubspec in $(find . -name pubspec.yaml -not -path '*/.*'); do
+          (cd "$(dirname "$pubspec")" && flutter pub get)
+        done
         dart format .
+        git clean -fdq # formatting never creates files, so drop generated pub and Flutter metadata
+        git ls-files -z '*pubspec.lock' | xargs -0 -r git checkout --
       shell: bash
       continue-on-error: true # Important so that commit step can pick up changes
 
--- a/action.yml
+++ b/action.yml
@@ -242,9 +242,12 @@ runs:
     # language version and silently falls back to the pre-3.7 short style.
     - name: Setup Flutter SDK
       if: github.event_name == 'pull_request' && inputs.dart == 'true' && (github.event.action == 'opened' || github.event.action == 'synchronize')
-      uses: subosito/flutter-action@v2
-      with:
-        channel: stable
+      run: |
+        echo "::group::Setup Flutter SDK"
+        trap 'echo "::endgroup::"' EXIT
+        git clone --depth 1 --branch stable https://github.com/flutter/flutter.git "$RUNNER_TEMP/flutter"
+        echo "$RUNNER_TEMP/flutter/bin" >> "$GITHUB_PATH"
+      shell: bash
 
     - name: Run Dart Formatter
       if: github.event_name == 'pull_request' && inputs.dart == 'true' && (github.event.action == 'opened' || github.event.action == 'synchronize')
--- a/action.yml
+++ b/action.yml
@@ -259,7 +259,7 @@ runs:
           (cd "$(dirname "$pubspec")" && flutter pub get) || ok=false
         done < <(find . -name pubspec.yaml -not -path '*/.*' -print0)
         $ok && dart format . || ok=false # never format with an unresolved language version
-        git clean -fdqx # formatting never creates files, so drop generated pub and Flutter metadata
+        git clean -fdq # formatting never creates files, so drop generated pub and Flutter metadata
         git ls-files -z '*pubspec.lock' | xargs -0 -r git checkout --
         $ok
       shell: bash
--- a/action.yml
+++ b/action.yml
@@ -255,9 +255,9 @@ runs:
         echo "::group::Run Dart Formatter"
         # Formatting never creates files: always drop generated pub/Flutter metadata and restore tracked lockfiles
         trap 'git clean -fdqx; git ls-files -z "*pubspec.lock" | xargs -0 -r git checkout --; echo "::endgroup::"' EXIT
-        for pubspec in $(find . -name pubspec.yaml -not -path '*/.*'); do
+        while IFS= read -r -d '' pubspec; do
           (cd "$(dirname "$pubspec")" && flutter pub get)
-        done
+        done < <(find . -name pubspec.yaml -not -path '*/.*' -print0)
         dart format .
       shell: bash
       continue-on-error: true # Important so that commit step can pick up changes
--- a/action.yml
+++ b/action.yml
@@ -253,13 +253,12 @@ runs:
       if: github.event_name == 'pull_request' && inputs.dart == 'true' && (github.event.action == 'opened' || github.event.action == 'synchronize')
       run: |
         echo "::group::Run Dart Formatter"
-        trap 'echo "::endgroup::"' EXIT
+        # Formatting never creates files: always drop generated pub/Flutter metadata and restore tracked lockfiles
+        trap 'git clean -fdqx; git ls-files -z "*pubspec.lock" | xargs -0 -r git checkout --; echo "::endgroup::"' EXIT
         for pubspec in $(find . -name pubspec.yaml -not -path '*/.*'); do
           (cd "$(dirname "$pubspec")" && flutter pub get)
         done
         dart format .
-        git clean -fdq # formatting never creates files, so drop generated pub and Flutter metadata
-        git ls-files -z '*pubspec.lock' | xargs -0 -r git checkout --
       shell: bash
       continue-on-error: true # Important so that commit step can pick up changes
 
--- a/action.yml
+++ b/action.yml
@@ -238,17 +238,20 @@ runs:
       continue-on-error: true
 
     # Dart formatting --------------------------------------------------------------------------------------------------
-    - name: Setup Dart SDK
+    # Flutter bundles Dart and resolves Flutter packages; without `pub get`, `dart format` cannot read the package
+    # language version and silently falls back to the pre-3.7 short style.
+    - name: Setup Flutter SDK
       if: github.event_name == 'pull_request' && inputs.dart == 'true' && (github.event.action == 'opened' || github.event.action == 'synchronize')
-      uses: dart-lang/setup-dart@v1
+      uses: subosito/flutter-action@v2
       with:
-        sdk: stable
+        channel: stable
 
     - name: Run Dart Formatter
       if: github.event_name == 'pull_request' && inputs.dart == 'true' && (github.event.action == 'opened' || github.event.action == 'synchronize')
       run: |
         echo "::group::Run Dart Formatter"
         trap 'echo "::endgroup::"' EXIT
+        find . -name pubspec.yaml -not -path '*/.*' -execdir flutter pub get \;
         dart format .
       shell: bash
       continue-on-error: true # Important so that commit step can pick up changes


UltralyticsAssistant:
👋 Hello @glenn-jocher, thank you for submitting a `ultralytics/actions` 🚀 PR! This automated message confirms your contribution was received, and an Ultralytics engineer will assist with the review. To ensure a seamless integration of your work, please review the following checklist:

action.yml
```diff
@@ -238,17 +238,20 @@ runs:
...
       with:
-        sdk: stable
+        channel: stable
 
     - name: Run Dart Formatter
       if: github.event_name == 'pull_request' && inputs.dart == 'true' && (github.event.action == 'opened' || github.event.action == 'synchronize')
       run: |
         echo "::group::Run Dart Formatter"
         trap 'echo "::endgroup::"' EXIT
+        find . -name pubspec.yaml -not -path '*/.*' -execdir flutter pub get \;
```
UltralyticsAssistant: 💡 **MEDIUM**: On a repository that does not already ignore Dart's generated state, this command creates `.dart_tool/package_config.json` and may create or update `pubspec.lock`; the existing commit step then runs `git add .` and pushes every staged change. A formatting-only run can therefore commit generated dependency metadata or library lockfiles. Keep these resolution artifacts out of the auto-format commit while retaining them during `dart format`.

UltralyticsAssistant (commented):
## 🔍 PR Review

Reviewed the complete action.yml diff. The Flutter setup and package resolution flow address the reported formatter issue, but generated pub artifacts can be accidentally committed by the existing blanket staging step in repositories without matching ignore rules.

<details><summary>💬 Posted 1 inline comment</summary>

- 💡 **MEDIUM** `action.yml:254` On a repository that does not already ignore Dart's generated state, this command creates `.dart_tool/package_config.json` and may create or update `pubspec.lock`; the existing commit step then runs `git add .` and pushes every staged change. A formatting-only run can therefore commit generated dependency metadata or library lockfiles. Keep these resolution artifacts out of the auto-format commit while retaining them during `dart format`.
</details>

action.yml
glenn-jocher: Not applicable in practice: `.dart_tool/` is gitignored by every `dart create`/`flutter create` template and by ultralytics/yolo-flutter-app (which also ignores `pubspec.lock` as a library should), so `git add .` picks up nothing from resolution. Where a repository does track `pubspec.lock`, `pub get` only rewrites it when it is already stale relative to `pubspec.yaml`, which is the same intentional behaviour `bun_update` has for Bun lockfiles. Left unchanged.

UltralyticsAssistant (commented):
## 🔍 PR Review 2

Since the previous review, the generated-artifact concern was discussed and left unchanged. The Flutter setup and package-resolution approach otherwise address the reported formatting issue, but the new resolution command can prevent all Dart formatting when any package fails to resolve. One medium-risk fail-fast issue remains.

<details><summary>💬 Posted 1 inline comment</summary>

- 💡 **MEDIUM** `action.yml:257` If any package has an unavailable dependency, incompatible SDK constraint, or transient pub.dev failure, this `find` command returns non-zero and GitHub's fail-fast Bash shell exits the step before `dart format .` runs at all. `continue-on-error` only allows later workflow steps after the step has failed; it does not continue commands within this script. Run all resolutions while tolerating per-package failures, then invoke the formatter, for example by handling the `find` status (`... \; || tr…
</details>

action.yml
```diff
@@ -238,17 +238,23 @@ runs:
...
+        git clone --depth 1 --branch stable https://github.com/flutter/flutter.git "$RUNNER_TEMP/flutter"
+        echo "$RUNNER_TEMP/flutter/bin" >> "$GITHUB_PATH"
+      shell: bash
 
     - name: Run Dart Formatter
       if: github.event_name == 'pull_request' && inputs.dart == 'true' && (github.event.action == 'opened' || github.event.action == 'synchronize')
       run: |
         echo "::group::Run Dart Formatter"
         trap 'echo "::endgroup::"' EXIT
+        find . -name pubspec.yaml -not -path '*/.*' -execdir flutter pub get \;
```
glenn-jocher: Not reproducible: `find`'s exit status reflects errors in `find` itself, not the exit status of `-execdir` commands. Verified on both runner families — `bash -e -c 'find . -name pubspec.yaml -execdir false \; ; echo $?'` prints 0 on macOS (BSD find) and on ubuntu:latest (GNU find) — so a failing `pub get` in one package never aborts the step, and `dart format .` still runs for every package. Left unchanged.

UltralyticsAssistant (commented):
## 🔍 PR Review 3

Since the last review, the Flutter SDK setup and package-resolution flow remain in place and address the reported Dart formatting regression. The per-package failure concern is settled by the maintainer's reproducibility reply. One previously identified generated-artifact risk remains because the resolution output is still included by the blanket staging step.

<details><summary>💬 Posted 1 inline comment</summary>

- 💡 **MEDIUM** `action.yml:257` In a repository that does not ignore Dart's generated state, `flutter pub get` creates `.dart_tool/package_config.json` and may create or update `pubspec.lock`. The later unchanged `git add .` stages and pushes those files, so a formatting-only run can commit dependency metadata. Exclude these resolution artifacts from the auto-format commit while retaining them for `dart format`.
</details>

action.yml
glenn-jocher: Added a `.git/info/exclude` entry for `.dart_tool/` so resolution output can never reach the auto-format commit regardless of the repository's .gitignore. `pubspec.lock` is only rewritten when it is already stale, which is intended (same as `bun_update`).

action.yml
```diff
@@ -238,17 +238,24 @@ runs:
...
+        echo "$RUNNER_TEMP/flutter/bin" >> "$GITHUB_PATH"
+      shell: bash
 
     - name: Run Dart Formatter
       if: github.event_name == 'pull_request' && inputs.dart == 'true' && (github.event.action == 'opened' || github.event.action == 'synchronize')
       run: |
         echo "::group::Run Dart Formatter"
         trap 'echo "::endgroup::"' EXIT
+        echo ".dart_tool/" >> .git/info/exclude # keep resolution artifacts out of the auto-format commit
+        find . -name pubspec.yaml -not -path '*/.*' -execdir flutter pub get \;
```
UltralyticsAssistant: ⚠️ **HIGH**: For a root-level package, `find .` evaluates the path as `./pubspec.yaml`, which also matches the `*/.*` pattern, so this predicate excludes the repository's primary `pubspec.yaml`. As a result, `flutter pub get` is skipped for the common single-package layout and `dart format .` still lacks the package config needed for the correct language-version style. Exclude hidden directories without excluding the root file, for example by changing the predicate to `-not -path '*/.*/*'`. ([gnu.org](https://www.gnu.org/software/findutils/manual/html_mono/find.html?utm_source=openai))

**Suggested change:**
```suggestion
        find . -name pubspec.yaml -not -path '*/.*/*' -execdir flutter pub get \;
```

UltralyticsAssistant (commented):
## 🔍 PR Review 4

Since the previous review, `.git/info/exclude` addresses the generated `.dart_tool` artifact concern, and the maintainer’s reproducibility reply settles the per-package failure-status concern. However, the new path filter skips a root-level `pubspec.yaml`, so the intended package-resolution fix still fails for standard single-package Flutter repositories.

<details><summary>💬 Posted 1 inline comment</summary>

- ⚠️ **HIGH** `action.yml:258` For a root-level package, `find .` evaluates the path as `./pubspec.yaml`, which also matches the `*/.*` pattern, so this predicate excludes the repository's primary `pubspec.yaml`. As a result, `flutter pub get` is skipped for the common single-package layout and `dart format .` still lacks the package config needed for the correct language-version style. Exclude hidden directories without excluding the root file, for example by changing the predicate to `-not -path '*/.*/*'`. ([gnu.org](https…
</details>

action.yml
glenn-jocher: Incorrect: `*/.*` requires a `/` immediately followed by `.`, and `./pubspec.yaml` has `p` after its only slash, so it is not excluded. Verified on both runner families: `find . -name pubspec.yaml -not -path '*/.*'` prints `./pubspec.yaml` and `./example/pubspec.yaml` in ultralytics/yolo-flutter-app (BSD find, macOS) and prints `./pubspec.yaml` while skipping `.hidden/pubspec.yaml` on ubuntu:latest (GNU find). Left unchanged.

UltralyticsAssistant (dismissed):
## 🔍 PR Review 5

Since review 4, the Flutter SDK setup and per-package `flutter pub get` flow remain unchanged; the previously raised concerns were addressed or settled in the discussion. I found no additional concrete defects in the current `action.yml` diff. LGTM.

UltralyticsAssistant (commented):
## 🔍 PR Review 6

Since review 5, the normal successful path remains correct and the prior root-package filtering concern is addressed. However, the current shell loop is still fail-fast on package-resolution errors, is unsafe for package paths containing whitespace, and can leave generated lockfiles staged when formatting fails.

<details><summary>💬 Posted 3 inline comments</summary>

- 💡 **MEDIUM** `action.yml:258` A nonzero `flutter pub get` exits this entire `bash` step before `dart format .`; GitHub runs explicit `shell: bash` steps with `bash --noprofile --norc -eo pipefail`, and `continue-on-error` only applies after the step exits. A package with an incompatible SDK, unavailable dependency, or transient resolution failure therefore prevents formatting for the whole repository. Tolerate each package-resolution failure and continue to the formatter.
- 💡 **MEDIUM** `action.yml:261` These cleanup commands are skipped whenever the preceding `dart format .` fails. Since `flutter pub get` can create `pubspec.lock`, an application package without a tracked lockfile (or a PR that deletes it) can leave an untracked generated lockfile that the later `git add .` commits. Run cleanup in a trap/finally path or explicitly remove/restore these artifacts even when formatting fails.
- 📝 **LOW** `action.yml:257` This command substitution performs shell word splitting, so a valid package path such as `packages/my app/pubspec.yaml` becomes multiple loop items. `dirname` then targets the wrong directory and that package is not resolved. Iterate over `find -print0` with a null-delimited read instead.
</details>

UltralyticsAssistant (commented):
## 🔍 PR Review 7

Since review 6, the shown diff leaves the package-resolution loop unchanged. The EXIT trap does run cleanup when formatting fails, but the direct loop still has two concrete issues: one failed `flutter pub get` can skip all Dart formatting, and paths containing whitespace are split incorrectly. Not LGTM.

- 💡 **MEDIUM** `action.yml:259` If any package has an incompatible SDK, unavailable dependency, or transient pub.dev failure, this direct `flutter pub get` command returns non-zero and the fail-fast Bash step exits before reaching `dart format .`. `continue-on-error` only applies after the whole step exits. Tolerate or record each package-resolution failure so formatting still runs for the remaining repository.
- 📝 **LOW** `action.yml:258` The command substitution performs shell word splitting, so a valid package path such as `packages/my app/pubspec.yaml` becomes multiple loop items and is not resolved correctly. Iterate over `find -print0` with a null-delimited `read` loop instead.
</details>

UltralyticsAssistant (commented):
## 🔍 PR Review 8

Since review 7, the package loop is now null-delimited and per-package resolution failures no longer abort the loop. The current diff still has two newly introduced issues: GNU-only `xargs` breaks the cleanup path on macOS, and the broad `git clean -fdqx` can irreversibly delete unrelated workspace artifacts. Not LGTM until cleanup is portable and narrowly scoped.

- ⚠️ **HIGH** `action.yml:262` `git clean -fdqx` deletes every ignored, untracked file and directory in the caller's workspace, not only Flutter/pub artifacts. If the workflow has already created `node_modules`, a virtualenv, generated assets, build outputs, or other ignored files, this formatting action irreversibly removes them and can break later steps. Replace this broad cleanup with targeted removal/restoration of only artifacts created by the package-resolution step.
- 💡 **MEDIUM** `action.yml:263` `xargs -r` is a GNU extension and is not supported by BSD/macOS `xargs`; the macOS synopsis omits this option. ([manpagez.com](https://www.manpagez.com/man/1/xargs/?utm_source=openai)) On macOS runners this command therefore fails before the final status check, leaving the formatter step failed and potentially leaving tracked `pubspec.lock` changes from `flutter pub get` unrestored. Use a portable null-delimited Bash loop or otherwise avoid the GNU-only flag.
</details>

action.yml
```diff
@@ -238,18 +238,30 @@ runs:
...
       run: |
         echo "::group::Run Dart Formatter"
         trap 'echo "::endgroup::"' EXIT
-        dart format .
+        ok=true
+        while IFS= read -r -d '' pubspec; do
+          (cd "$(dirname "$pubspec")" && flutter pub get) || ok=false
+        done < <(find . -name pubspec.yaml -not -path '*/.*' -print0)
+        $ok && dart format . || ok=false # never format with an unresolved language version
+        git clean -fdqx # formatting never creates files, so drop generated pub and Flutter metadata
```
glenn-jocher: Narrowed to `git clean -fdq` (untracked, non-ignored only): ignored files are already excluded from the commit step's `git add .`, so only un-ignored resolution artifacts need removing.

UltralyticsAssistant (commented):
## 🔍 PR Review 9

Since review 8, the package loop is null-delimited and per-package resolution failures no longer abort the loop, and the cleanup no longer removes ignored files. One remaining data-loss risk is the repository-wide deletion of non-ignored untracked content by `git clean -fdq`; the cleanup should be scoped to artifacts created by package resolution. Not LGTM.

<details><summary>💬 Posted 1 inline comment</summary>

- ⚠️ **HIGH** `action.yml:262` `git clean -fdq` still recursively deletes every non-ignored untracked file and directory in the caller's workspace, not only artifacts created by `flutter pub get`. For example, a workflow that generates an untracked fixture, build output, or metadata before invoking this composite action will lose it irreversibly, and the later `git add .` cannot recover it. Track the files created by package resolution and remove only those paths instead of cleaning the whole repository.
</details>

action.yml
```diff
@@ -238,18 +238,30 @@ runs:
...
       run: |
         echo "::group::Run Dart Formatter"
         trap 'echo "::endgroup::"' EXIT
-        dart format .
+        ok=true
+        while IFS= read -r -d '' pubspec; do
+          (cd "$(dirname "$pubspec")" && flutter pub get) || ok=false
+        done < <(find . -name pubspec.yaml -not -path '*/.*' -print0)
+        $ok && dart format . || ok=false # never format with an unresolved language version
+        git clean -fdq # formatting never creates files, so drop generated pub and Flutter metadata
```
UltralyticsAssistant, line 262: ⚠️ **HIGH**: `git clean -fdq` still recursively deletes every non-ignored untracked file and directory in the caller's workspace, not only artifacts created by `flutter pub get`. For example, a workflow that generates an untracked fixture, build output, or metadata before invoking this composite action will lose it irreversibly, and the later `git add .` cannot recover it. Track the files created by package resolution and remove only those paths instead of cleaning the whole repository.

UltralyticsAssistant:
🎉 PR [#876](https://github.com/ultralytics/actions/pull/876) has been merged—thank you, @glenn-jocher!

> “Alone we can do so little; together we can do so much.” — Helen Keller

By installing Flutter, resolving packages before formatting, and cleaning up generated metadata, this update makes Dart formatting reliably use the correct language version across relevant pull request workflows.

PR #879, opened 2026-08-17
Enforce per-attempt retry timeouts

glenn-jocher:
## Summary

- make the existing timeout_minutes input bound each retry attempt
- terminate the full child process tree when an attempt hangs
- verify timeout-and-retry behavior on Linux, macOS, and Windows
- bump ultralytics-actions from 0.3.9 to 0.3.10

## Motivation

The scheduled Ultralytics CI run https://github.com/ultralytics/ultralytics/actions/runs/31935546886 spent six hours in one hung pytest process. The retry wrapper only checked its timeout between completed attempts, so it could neither interrupt the hang nor reach its retry.

A local inventory found 53 uses across 13 repositories. Their explicit values (5-120 minutes) already match per-attempt workloads, while each workflow job remains the owner of its overall ceiling.

## Validation

- local process-tree timeout and retry simulation with a SIGTERM-ignoring descendant
- 166 package tests passed
- Ruff, Actionlint, Prettier, and diff checks passed
- cross-platform retry-action CI pending on the final head

## 🛠️ PR Summary

### 🌟 Summary

The retry action now enforces `timeout_minutes` separately for each attempt and terminates timed-out process trees so retries can proceed instead of waiting indefinitely.

### 📊 Key Changes

- Replaced the previous between-attempt timeout check with active per-attempt supervision.
- Added Linux and macOS timeout handling using Python process groups, descendant tracking, and termination signals.
- Added Windows timeout handling with PowerShell process discovery and `taskkill` process-tree termination.
- Updated `timeout_minutes` documentation and bumped the action version from `0.3.9` to `0.3.10`.
- Expanded retry tests across Ubuntu, macOS, and Windows, including hung commands, detached descendants, and Windows Python-shell execution.

### 🎯 Purpose & Impact

- A hung attempt is interrupted after its configured timeout, its tracked child processes are terminated, and the configured retry sequence can continue.
- `timeout_minutes` now limits each attempt rather than the combined retry sequence; retry delays are no longer capped by a shared timeout budget.
- Timeout supervision requires Python 3 on Linux and macOS, while detached macOS descendants that clear their inherited environment remain outside the documented containment boundary.

### DIFF
--- a/.github/workflows/test-retry.yml
+++ b/.github/workflows/test-retry.yml
@@ -82,7 +82,7 @@ jobs:
                 child_file=$(cygpath -w /tmp/retry_child)
                 powershell.exe -NoProfile -Command "[IO.File]::WriteAllText('$child_file', [string]\$PID); Start-Sleep 120" &
               else
-                python3 -c 'import os, signal, time; os.fork() and os._exit(0); os.setsid(); os.fork() and os._exit(0); open("/tmp/retry_child", "w").write(str(os.getpid())); signal.signal(signal.SIGTERM, signal.SIG_IGN); time.sleep(120)'
+                python3 -c 'import os, signal, time; os.fork() and os._exit(0); os.setsid(); os.fork() and os._exit(0); open("/tmp/retry_child", "w").write(str(os.getpid())); os.environ.clear(); signal.signal(signal.SIGTERM, signal.SIG_IGN); time.sleep(120)'
                 sleep 120 &
               fi
               wait $!
--- a/.github/workflows/test-retry.yml
+++ b/.github/workflows/test-retry.yml
@@ -12,7 +12,14 @@ on:
 
 jobs:
   test-bash:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    defaults:
+      run:
+        shell: bash
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest, windows-latest]
     steps:
       - uses: actions/checkout@v7
 
@@ -60,7 +67,7 @@ jobs:
         with:
           retries: 1
           retry_delay_seconds: 0
-          attempt_timeout_minutes: 1
+          timeout_minutes: 1
           run: |
             if [ ! -f /tmp/retry_attempt ]; then
               touch /tmp/retry_attempt
--- a/.github/workflows/test-retry.yml
+++ b/.github/workflows/test-retry.yml
@@ -79,14 +79,20 @@ jobs:
             if [ ! -f /tmp/retry_attempt ]; then
               touch /tmp/retry_attempt
               if [ "$RUNNER_OS" = "Windows" ]; then
-                bash -c 'trap "" TERM; sleep 120' &
+                child_file=$(cygpath -w /tmp/retry_child)
+                powershell.exe -NoProfile -Command "[IO.File]::WriteAllText('$child_file', [string]\$PID); Start-Sleep 120" &
               else
-                python3 -c 'import os, signal, time; os.setsid(); signal.signal(signal.SIGTERM, signal.SIG_IGN); time.sleep(120)' &
+                python3 -c 'import os, signal, time; os.fork() and os._exit(0); os.setsid(); os.fork() and os._exit(0); open("/tmp/retry_child", "w").write(str(os.getpid())); signal.signal(signal.SIGTERM, signal.SIG_IGN); time.sleep(120)'
+                sleep 120 &
               fi
-              echo $! > /tmp/retry_child
               wait $!
             fi
-            if [ "$RUNNER_OS" != "Windows" ] && kill -0 "$(cat /tmp/retry_child)" 2>/dev/null; then
+            if [ "$RUNNER_OS" = "Windows" ]; then
+              powershell.exe -NoProfile -Command "if (Get-Process -Id $(cat /tmp/retry_child) -ErrorAction SilentlyContinue) { exit 0 } else { exit 1 }" && child_alive=true
+            elif kill -0 "$(cat /tmp/retry_child)" 2>/dev/null; then
+              child_alive=true
+            fi
+            if [ "$child_alive" = true ]; then
               echo "::error::Timed-out child is still running"
               exit 1
             fi
--- a/.github/workflows/test-retry.yml
+++ b/.github/workflows/test-retry.yml
@@ -33,6 +33,13 @@ jobs:
             echo "This should succeed first try"
             true
 
+      - name: Test Python shell on Windows
+        if: matrix.os == 'windows-latest'
+        uses: ./retry
+        with:
+          shell: python
+          run: print("Python shell works")
+
       - name: Test retry with eventual success
         uses: ./retry
         env:
--- a/.github/workflows/test-retry.yml
+++ b/.github/workflows/test-retry.yml
@@ -86,7 +86,7 @@ jobs:
               echo $! > /tmp/retry_child
               wait $!
             fi
-            if kill -0 "$(cat /tmp/retry_child)" 2>/dev/null; then
+            if [ "$RUNNER_OS" != "Windows" ] && kill -0 "$(cat /tmp/retry_child)" 2>/dev/null; then
               echo "::error::Timed-out child is still running"
               exit 1
             fi
--- a/.github/workflows/test-retry.yml
+++ b/.github/workflows/test-retry.yml
@@ -71,7 +71,8 @@ jobs:
           run: |
             if [ ! -f /tmp/retry_attempt ]; then
               touch /tmp/retry_attempt
-              sleep 120
+              bash -c 'trap "" TERM; sleep 120' &
+              wait $!
             fi
             rm -f /tmp/retry_attempt
 
--- a/.github/workflows/test-retry.yml
+++ b/.github/workflows/test-retry.yml
@@ -55,6 +55,19 @@ jobs:
             rm -f /tmp/retry_attempt
             echo "Success on attempt $((count + 1))!"
 
+      - name: Test hung command retry
+        uses: ./retry
+        with:
+          retries: 1
+          retry_delay_seconds: 0
+          attempt_timeout_minutes: 1
+          run: |
+            if [ ! -f /tmp/retry_attempt ]; then
+              touch /tmp/retry_attempt
+              sleep 120
+            fi
+            rm -f /tmp/retry_attempt
+
       - name: Test multi-line failure
         uses: ./retry
         continue-on-error: true
--- a/.github/workflows/test-retry.yml
+++ b/.github/workflows/test-retry.yml
@@ -81,8 +81,11 @@ jobs:
               if [ "$RUNNER_OS" = "Windows" ]; then
                 child_file=$(cygpath -w /tmp/retry_child)
                 powershell.exe -NoProfile -Command "[IO.File]::WriteAllText('$child_file', [string]\$PID); Start-Sleep 120" &
-              else
+              elif [ "$RUNNER_OS" = "Linux" ]; then
                 python3 -c 'import os, signal, time; os.fork() and os._exit(0); os.setsid(); os.fork() and os._exit(0); open("/tmp/retry_child", "w").write(str(os.getpid())); os.environ.clear(); signal.signal(signal.SIGTERM, signal.SIG_IGN); time.sleep(120)'
+                sleep 120 &
+              else
+                python3 -c 'import os, signal, time; os.fork() and os._exit(0); os.setsid(); os.fork() and os._exit(0); open("/tmp/retry_child", "w").write(str(os.getpid())); signal.signal(signal.SIGTERM, signal.SIG_IGN); time.sleep(120)'
                 sleep 120 &
               fi
               wait $!
--- a/.github/workflows/test-retry.yml
+++ b/.github/workflows/test-retry.yml
@@ -24,7 +24,7 @@ jobs:
       - uses: actions/checkout@v7
 
       - name: Cleanup any leftover test files
-        run: rm -f /tmp/retry_attempt /tmp/retry_attempt_py
+        run: rm -f /tmp/retry_attempt /tmp/retry_attempt_py /tmp/retry_child
 
       - name: Test successful command
         uses: ./retry
@@ -78,10 +78,19 @@ jobs:
           run: |
             if [ ! -f /tmp/retry_attempt ]; then
               touch /tmp/retry_attempt
-              bash -c 'trap "" TERM; sleep 120' &
+              if [ "$RUNNER_OS" = "Windows" ]; then
+                bash -c 'trap "" TERM; sleep 120' &
+              else
+                python3 -c 'import os, signal, time; os.setsid(); signal.signal(signal.SIGTERM, signal.SIG_IGN); time.sleep(120)' &
+              fi
+              echo $! > /tmp/retry_child
               wait $!
             fi
-            rm -f /tmp/retry_attempt
+            if kill -0 "$(cat /tmp/retry_child)" 2>/dev/null; then
+              echo "::error::Timed-out child is still running"
+              exit 1
+            fi
+            rm -f /tmp/retry_attempt /tmp/retry_child
 
       - name: Test multi-line failure
         uses: ./retry
--- a/retry/action.yml
+++ b/retry/action.yml
@@ -142,6 +142,7 @@ runs:
             taskkill.exe /PID $process.Id /T /F | Out-Null
             exit 124
         }
+        $process.WaitForExit()
         exit $process.ExitCode
         '
             else
--- a/retry/action.yml
+++ b/retry/action.yml
@@ -126,7 +126,7 @@ runs:
                 echo "::group::Retry $((attempt - 1)) of $retries"
             fi
 
-            # Execute the script with a deadline, terminating its entire process tree on timeout.
+            # Execute the script with a deadline, terminating its process group and tracked descendants on timeout.
             timeout_marker="${TEMP_SCRIPT}.timeout"
             rm -f "$timeout_marker"
             if [ "$RETRY_RUNNER_OS" = "Windows" ]; then
--- a/retry/action.yml
+++ b/retry/action.yml
@@ -169,7 +169,7 @@ runs:
         def tracked_processes(token):
             """Return processes carrying the attempt's unique inherited environment marker."""
             marker = f"ULTRALYTICS_RETRY_TOKEN={token}"
-            processes = subprocess.check_output(["ps", "eww", "-axo", "pid=,command="], text=True)
+            processes = subprocess.check_output(["ps", "axeww", "-o", "pid=", "-o", "command="], text=True)
             return {int(line.split(None, 1)[0]) for line in processes.splitlines() if marker in line}
 
         def send_signal(process_group, process_ids, sig):
--- a/retry/action.yml
+++ b/retry/action.yml
@@ -16,8 +16,7 @@
 #         python setup.py install
 #         pytest tests/
 #       retries: 2               # Retry twice after initial attempt (3 total runs)
-#       timeout_minutes: 30      # Total timeout for all attempts combined
-#       attempt_timeout_minutes: 10  # Maximum time for each attempt (0 disables)
+#       timeout_minutes: 30      # Maximum time for each attempt
 #       retry_delay_seconds: 10  # Base delay between retries in seconds
 #       backoff: exponential     # exponential (10s, 20s, 40s, ...) or fixed
 #       jitter: true             # Randomize delay to 80-120% to avoid thundering herd
@@ -27,23 +26,19 @@ name: "Step-Level Retry"
 description: "Retries a step while preserving its full context"
 inputs:
   timeout_minutes:
-    description: "Maximum total time in minutes for all attempts (checked between attempts; does not interrupt a running attempt)"
+    description: "Maximum time in minutes for each attempt"
     required: false
     default: "360"
   retries:
     description: "Number of retry attempts after initial run"
     required: false
     default: "3"
-  attempt_timeout_minutes:
-    description: "Maximum time in minutes for each attempt; 0 disables the per-attempt timeout"
-    required: false
-    default: "0"
   retry_delay_seconds:
     description: "Base delay between retries in seconds"
     required: false
     default: "10"
   backoff:
-    description: "Backoff strategy: exponential (base * 2^n) or fixed. Delays are bounded by timeout_minutes."
+    description: "Backoff strategy: exponential (base * 2^n) or fixed"
     required: false
     default: "exponential"
   jitter:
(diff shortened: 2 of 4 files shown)


UltralyticsAssistant:
👋 Hello @glenn-jocher, thank you for submitting a `ultralytics/actions` 🚀 PR! This automated message confirms your contribution was received, and an Ultralytics engineer will assist with the review. To ensure a seamless integration of your work, please review the following checklist:

UltralyticsAssistant (commented):
## 🔍 PR Review

Reviewed the retry action implementation, documentation, and CI coverage. The per-attempt timeout is wired through correctly, but POSIX cleanup has process-tree and process-group races that can leave descendants running or misreport a timeout.

- ⚠️ **HIGH** `retry/action.yml:169` `process.wait(timeout=5)` only waits for the shell/Python root process, not its descendants. If a child ignores SIGTERM, the root can exit while that child remains alive; this returns from `wait()` and skips the SIGKILL, so the retry starts with the original process still running and potentially mutating shared files or consuming resources. Force-kill the process group after the grace period independently of the root's exit, while tolerating an already-gone group.
- 📝 **LOW** `retry/action.yml:167` Both `os.killpg` calls have an uncaught race: the timed-out process group may exit between `process.wait(timeout=...)` raising and this call. In that case `killpg` raises `ProcessLookupError`, causing the wrapper to return a generic status 1 instead of the intended timeout status 124; with no retries left, this misreports a command that finished at the deadline. Handle an already-gone process group explicitly.
</details>

retry/action.yml
```diff
@@ -134,14 +141,47 @@ runs:
...
+                subprocess.run(
+                    ["taskkill", "/PID", str(process.pid), "/T", "/F"],
+                    stdout=subprocess.DEVNULL,
+                    stderr=subprocess.DEVNULL,
+                    check=False,
+                )
+            else:
+                os.killpg(process.pid, signal.SIGTERM)
+                try:
+                    process.wait(timeout=5)
```
UltralyticsAssistant: ⚠️ **HIGH**: `process.wait(timeout=5)` only waits for the shell/Python root process, not its descendants. If a child ignores SIGTERM, the root can exit while that child remains alive; this returns from `wait()` and skips the SIGKILL, so the retry starts with the original process still running and potentially mutating shared files or consuming resources. Force-kill the process group after the grace period independently of the root's exit, while tolerating an already-gone group.

retry/action.yml
```diff
@@ -134,14 +141,47 @@ runs:
...
+        except subprocess.TimeoutExpired:
+            if os.name == "nt":
+                subprocess.run(
+                    ["taskkill", "/PID", str(process.pid), "/T", "/F"],
+                    stdout=subprocess.DEVNULL,
+                    stderr=subprocess.DEVNULL,
+                    check=False,
+                )
+            else:
+                os.killpg(process.pid, signal.SIGTERM)
```
UltralyticsAssistant: 📝 **LOW**: Both `os.killpg` calls have an uncaught race: the timed-out process group may exit between `process.wait(timeout=...)` raising and this call. In that case `killpg` raises `ProcessLookupError`, causing the wrapper to return a generic status 1 instead of the intended timeout status 124; with no retries left, this misreports a command that finished at the deadline. Handle an already-gone process group explicitly.

retry/action.yml
```diff
@@ -126,22 +126,72 @@ runs:
...
+        import signal
+        import subprocess
+        import sys
+        import time
+
+        shell_type, script, timeout = sys.argv[1], sys.argv[2], int(sys.argv[3])
+        command = [sys.executable, script] if shell_type == "python" else ["bash", "-e", script]
+        process = subprocess.Popen(command, start_new_session=True)
+        try:
+            sys.exit(process.wait(timeout=timeout))
```
UltralyticsAssistant: 📝 **LOW**: When the child exits due to a signal, `Popen.wait()` returns a negative status (for example, `-15` for SIGTERM), and passing it directly to `sys.exit()` makes the helper report `241` instead of the conventional shell status `143`. This changes the observable exit code from the previous direct `bash -e` path and can break workflows that distinguish signal failures. Convert negative return codes to `128 + signal_number` before exiting; Python documents negative `Popen.returncode` values for signal termination, while Bash uses `128+N` for the corresponding status. ([docs.python.org](https://docs.python.org/3/library/subprocess.html?highlight=universal_newlines&utm_source=openai))

retry/action.yml
```diff
@@ -126,22 +126,72 @@ runs:
...
+                    os.killpg(process.pid, signal.SIGKILL)
+                except ProcessLookupError:
+                    pass
+            process.wait()
+            sys.exit(124)
+        PY
             fi
             exit_code=$?
 
+            if [ "$exit_code" -eq 124 ]; then
```
UltralyticsAssistant: 📝 **LOW**: Exit status `124` is not reserved by the child command, so a user script that explicitly returns `124` is treated and reported as a wrapper timeout even when it completed before the deadline. This produces a misleading timeout annotation and obscures the actual failure. Track timeout state separately from the child exit code instead of inferring it solely from `$exit_code`.

UltralyticsAssistant (commented):
## 🔍 PR Review 2

Review 2: the earlier POSIX process-group findings are addressed—the grace-period cleanup now checks descendants independently and tolerates groups disappearing during cleanup. Two status-handling issues remain: signal-terminated commands change exit-code conventions on POSIX, and ordinary exit code 124 is misreported as a timeout.

- 📝 **LOW** `retry/action.yml:167` When the child exits due to a signal, `Popen.wait()` returns a negative status (for example, `-15` for SIGTERM), and passing it directly to `sys.exit()` makes the helper report `241` instead of the conventional shell status `143`. This changes the observable exit code from the previous direct `bash -e` path and can break workflows that distinguish signal failures. Convert negative return codes to `128 + signal_number` before exiting; Python documents negative `Popen.returncode` values for signa…
- 📝 **LOW** `retry/action.yml:191` Exit status `124` is not reserved by the child command, so a user script that explicitly returns `124` is treated and reported as a wrapper timeout even when it completed before the deadline. This produces a misleading timeout annotation and obscures the actual failure. Track timeout state separately from the child exit code instead of inferring it solely from `$exit_code`.
</details>

retry/action.yml
```diff
@@ -126,22 +126,78 @@ runs:
...
+            if [ "$RETRY_RUNNER_OS" = "Windows" ]; then
+                RETRY_EXECUTABLE=$(cygpath -w "$BASH") \
+                RETRY_SCRIPT=$(cygpath -w "$TEMP_SCRIPT") \
+                RETRY_SHELL_TYPE="$shell_type" \
+                RETRY_TIMEOUT_MARKER=$(cygpath -w "$timeout_marker") \
+                RETRY_TIMEOUT_SECONDS="$timeout_seconds" \
+                powershell.exe -NoProfile -Command '
+        $startInfo = New-Object System.Diagnostics.ProcessStartInfo
+        if ($env:RETRY_SHELL_TYPE -eq "python") {
+            $startInfo.FileName = "py.exe"
```
UltralyticsAssistant: 💡 **MEDIUM**: On Windows, `shell: python` now bypasses the interpreter resolved by the caller and always invokes `py.exe -3`. The previous path ran `python3 "$TEMP_SCRIPT"`; a job that prepends a virtualenv or `setup-python` interpreter to `PATH` can now execute with a different Python installation and fail on missing packages or the wrong version. Python's launcher prefers an active virtual environment only when no version is explicitly requested, while `-3` selects a Python 3 runtime. ([docs.python.org](https://docs.python.org/3/using/windows.html?highlight=launcher%23python-launcher-for-windows&utm_source=openai)) Resolve the same Python executable in Bash and pass its absolute path into PowerShell instead of hard-coding `py.exe -3`.

retry/action.yml
```diff
@@ -126,22 +126,78 @@ runs:
...
+        shell_type, script, timeout, timeout_marker = sys.argv[1], sys.argv[2], int(sys.argv[3]), sys.argv[4]
+        command = [sys.executable, script] if shell_type == "python" else ["bash", "-e", script]
+        process = subprocess.Popen(command, start_new_session=True)
+        try:
+            return_code = process.wait(timeout=timeout)
+            sys.exit(128 - return_code if return_code < 0 else return_code)
+        except subprocess.TimeoutExpired:
+            open(timeout_marker, "w").close()
+            try:
+                os.killpg(process.pid, signal.SIGTERM)
```
UltralyticsAssistant: 💡 **MEDIUM**: The POSIX cleanup only signals the original process group. A descendant can call `setsid` (for example, a daemonized subprocess) and move into a new session/process group; the helper then exits with 124 and the next retry starts while that descendant is still running. `setsid` explicitly creates a new session and process group. ([man7.org](https://man7.org/linux/man-pages/man2/setsid.2.html?utm_source=openai)) Use a supervisor or descendant-tracking mechanism that also covers children which create new groups, or avoid claiming that all process trees are terminated.

UltralyticsAssistant (commented):
## 🔍 PR Review 3

Since review 2, the signal exit-code conversion and separate timeout marker address the previously reported status-handling issues. The remaining diff introduces a Windows Python-interpreter compatibility regression and does not reliably terminate POSIX descendants that detach into a new process group, so this is not LGTM yet.

- 💡 **MEDIUM** `retry/action.yml:141` On Windows, `shell: python` now bypasses the interpreter resolved by the caller and always invokes `py.exe -3`. The previous path ran `python3 "$TEMP_SCRIPT"`; a job that prepends a virtualenv or `setup-python` interpreter to `PATH` can now execute with a different Python installation and fail on missing packages or the wrong version. Python's launcher prefers an active virtual environment only when no version is explicitly requested, while `-3` selects a Python 3 runtime. ([docs.python.org](ht…
- 💡 **MEDIUM** `retry/action.yml:176` The POSIX cleanup only signals the original process group. A descendant can call `setsid` (for example, a daemonized subprocess) and move into a new session/process group; the helper then exits with 124 and the next retry starts while that descendant is still running. `setsid` explicitly creates a new session and process group. ([man7.org](https://man7.org/linux/man-pages/man2/setsid.2.html?utm_source=openai)) Use a supervisor or descendant-tracking mechanism that also covers children which cre…
</details>

retry/action.yml
```diff
@@ -126,22 +126,105 @@ runs:
...
+
+        shell_type, script, timeout, timeout_marker = sys.argv[1], sys.argv[2], int(sys.argv[3]), sys.argv[4]
+        command = [sys.executable, script] if shell_type == "python" else ["bash", "-e", script]
+        process = subprocess.Popen(command, start_new_session=True)
+        try:
+            return_code = process.wait(timeout=timeout)
+            sys.exit(128 - return_code if return_code < 0 else return_code)
+        except subprocess.TimeoutExpired:
+            open(timeout_marker, "w").close()
+            process_ids = descendants(process.pid)
```
UltralyticsAssistant: 💡 **MEDIUM**: The detached-descendant issue from review 3 is still present: `descendants()` takes a one-time PPID snapshot. A command can double-fork or daemonize before the deadline, causing the surviving process to be reparented before this scan; it is then absent from `process_ids` and, if it created a new session, unreachable through `killpg(process.pid, ...)`. The helper returns 124 and starts the retry while that process continues running. Use a supervisor/cgroup-based cleanup mechanism or otherwise retain ownership until descendants terminate; this snapshot cannot guarantee the documented full process-tree cleanup.

.github/workflows/test-retry.yml
```diff
@@ -55,6 +69,29 @@ jobs:
...
+              touch /tmp/retry_attempt
+              if [ "$RUNNER_OS" = "Windows" ]; then
+                bash -c 'trap "" TERM; sleep 120' &
+              else
+                python3 -c 'import os, signal, time; os.setsid(); signal.signal(signal.SIGTERM, signal.SIG_IGN); time.sleep(120)' &
+              fi
+              echo $! > /tmp/retry_child
+              wait $!
+            fi
+            if [ "$RUNNER_OS" != "Windows" ] && kill -0 "$(cat /tmp/retry_child)" 2>/dev/null; then
```
UltralyticsAssistant: 📝 **LOW**: The Windows path skips the only assertion that the timed-out child was terminated. If `taskkill /T` misses or fails to remove the child, the retry still succeeds because the PID check is gated to non-Windows, so CI passes while leaking the 120-second process. Add a Windows PID-liveness check before deleting `/tmp/retry_child` (or otherwise make the retry fail when that child remains alive).

UltralyticsAssistant (commented):
## 🔍 PR Review 4

Review 4: the earlier interpreter and exit-status issues remain addressed. The per-attempt timeout is wired correctly, but POSIX descendants can still escape after reparenting, and the new Windows test does not verify child cleanup. These are the remaining risks; not LGTM yet.

- 💡 **MEDIUM** `retry/action.yml:202` The detached-descendant issue from review 3 is still present: `descendants()` takes a one-time PPID snapshot. A command can double-fork or daemonize before the deadline, causing the surviving process to be reparented before this scan; it is then absent from `process_ids` and, if it created a new session, unreachable through `killpg(process.pid, ...)`. The helper returns 124 and starts the retry while that process continues running. Use a supervisor/cgroup-based cleanup mechanism or otherwise re…
- 📝 **LOW** `.github/workflows/test-retry.yml:89` The Windows path skips the only assertion that the timed-out child was terminated. If `taskkill /T` misses or fails to remove the child, the retry still succeeds because the PID check is gated to non-Windows, so CI passes while leaking the 120-second process. Add a Windows PID-liveness check before deleting `/tmp/retry_child` (or otherwise make the retry fail when that child remains alive).
</details>

UltralyticsAssistant (commented):
## 🔍 PR Review 5

Since review 4, the PPID snapshot has been replaced with global token tracking, the Windows liveness assertion is now present, and the prior exit-status issues remain addressed. Two risks remain: the POSIX implementation adds an undocumented Python dependency to Bash-only users, and descendants can still escape cleanup by clearing the tracking environment before detaching.

- 💡 **MEDIUM** `retry/action.yml:161` The new POSIX timeout path unconditionally invokes `python3` even for the default `shell: bash`; previously, Bash retries did not require Python. On a self-hosted Linux/macOS runner with Bash but no `python3`, every attempt now fails in the supervisor with status 127 instead of running the user command. Keep the Bash path free of this undeclared dependency, or explicitly validate and document Python as a new prerequisite.
- 💡 **MEDIUM** `retry/action.yml:191` Tracking descendants through a mutable inherited environment variable still leaves an escape path. For example, `setsid env -i ...` can clear `ULTRALYTICS_RETRY_TOKEN` before creating a new session/process group; `tracked_processes()` then omits that process, while `killpg(process.pid, ...)` cannot reach it. The wrapper returns 124 and starts the retry with the detached process still running. Use an OS-owned supervisor/cgroup/job object or another mechanism that the child cannot remove. ([man7.…
</details>

retry/action.yml
```diff
@@ -126,22 +126,96 @@ runs:
...
+            $startInfo.FileName = $env:RETRY_EXECUTABLE
+            $startInfo.Arguments = "-e `"$env:RETRY_SCRIPT`""
+        }
+        $startInfo.UseShellExecute = $false
+        $process = New-Object System.Diagnostics.Process
+        $process.StartInfo = $startInfo
+        [void]$process.Start()
+        if (-not $process.WaitForExit([int]$env:RETRY_TIMEOUT_SECONDS * 1000)) {
+            [System.IO.File]::WriteAllText($env:RETRY_TIMEOUT_MARKER, "")
+            taskkill.exe /PID $process.Id /T /F | Out-Null
```
UltralyticsAssistant: 💡 **MEDIUM**: There is still a Windows cleanup race: if the direct process exits after `WaitForExit` returns false but before `taskkill.exe` runs—for example, after leaving a detached child—the `/PID ... /T` operation is rooted at a PID that no longer exists, so the child can survive while this wrapper unconditionally reports a timeout and retries. Microsoft documents `/T` as terminating the specified process and its children; retain descendants through an OS-owned job/supervisor or otherwise verify and complete cleanup when the root disappears. ([learn.microsoft.com](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/taskkill?utm_source=openai))

retry/action.yml
```diff
@@ -126,22 +126,96 @@ runs:
...
+            exit 124
+        }
+        $process.WaitForExit()
+        exit $process.ExitCode
+        '
             else
-                bash -e "$TEMP_SCRIPT"
+                python3 - "$shell_type" "$TEMP_SCRIPT" "$timeout_seconds" "$timeout_marker" <<'PY'
```
UltralyticsAssistant: 💡 **MEDIUM**: Every Bash attempt on Linux/macOS now starts through `python3`; the Windows branch also resolves `python3` for `shell: bash` at line 134. On a self-hosted runner with Bash but no Python 3, the wrapper exits 127 before executing the user command, regressing the 0.3.9 behavior. Keep Bash supervision independent of Python, or validate Python as a prerequisite on every platform and shell.

retry/action.yml
```diff
@@ -126,22 +126,96 @@ runs:
...
+                try:
+                    os.kill(pid, sig)
+                except (PermissionError, ProcessLookupError):
+                    pass
+
+        shell_type, script, timeout, timeout_marker = sys.argv[1], sys.argv[2], int(sys.argv[3]), sys.argv[4]
+        command = [sys.executable, script] if shell_type == "python" else ["bash", "-e", script]
+        token = uuid.uuid4().hex
+        environment = os.environ.copy()
+        environment["ULTRALYTICS_RETRY_TOKEN"] = token
```
UltralyticsAssistant: 💡 **MEDIUM**: The marker is the only way to find descendants after they leave the original process group. A child can `setsid`/reparent and then clear `ULTRALYTICS_RETRY_TOKEN` (for example via `env -i`) before the deadline; it is then absent from `tracked_processes()` and unreachable through `killpg(process.pid, ...)`. The wrapper returns 124 and starts the next retry while that process continues running. Track descendants independently before detachment or use OS-owned containment rather than relying on a mutable inherited environment variable.

UltralyticsAssistant (commented):
## 🔍 PR Review 6

Since review 5, the prior limitations are documented but not behaviorally resolved, and the new Windows timeout path adds a root-process cleanup race. The version bump, per-attempt timeout wiring, and cross-platform test matrix otherwise look consistent, but this is not LGTM until timeout cleanup is reliable and the new Python dependency is handled without regressing Bash-only runners.

<details><summary>💬 Posted 3 inline comments</summary>

- 💡 **MEDIUM** `retry/action.yml:154` There is still a Windows cleanup race: if the direct process exits after `WaitForExit` returns false but before `taskkill.exe` runs—for example, after leaving a detached child—the `/PID ... /T` operation is rooted at a PID that no longer exists, so the child can survive while this wrapper unconditionally reports a timeout and retries. Microsoft documents `/T` as terminating the specified process and its children; retain descendants through an OS-owned job/supervisor or otherwise verify and comp…
- 💡 **MEDIUM** `retry/action.yml:161` Every Bash attempt on Linux/macOS now starts through `python3`; the Windows branch also resolves `python3` for `shell: bash` at line 134. On a self-hosted runner with Bash but no Python 3, the wrapper exits 127 before executing the user command, regressing the 0.3.9 behavior. Keep Bash supervision independent of Python, or validate Python as a prerequisite on every platform and shell.
- 💡 **MEDIUM** `retry/action.yml:191` The marker is the only way to find descendants after they leave the original process group. A child can `setsid`/reparent and then clear `ULTRALYTICS_RETRY_TOKEN` (for example via `env -i`) before the deadline; it is then absent from `tracked_processes()` and unreachable through `killpg(process.pid, ...)`. The wrapper returns 124 and starts the next retry while that process continues running. Track descendants independently before detachment or use OS-owned containment rather than relying on a…
</details>

retry/action.yml
```diff
@@ -126,22 +131,135 @@ runs:
...
+            $startInfo.FileName = $env:RETRY_EXECUTABLE
+            $startInfo.Arguments = "-e `"$env:RETRY_SCRIPT`""
+        }
+        $startInfo.UseShellExecute = $false
+        $process = New-Object System.Diagnostics.Process
+        $process.StartInfo = $startInfo
+        [void]$process.Start()
+        if (-not $process.WaitForExit([int]$env:RETRY_TIMEOUT_SECONDS * 1000)) {
+            [System.IO.File]::WriteAllText($env:RETRY_TIMEOUT_MARKER, "")
+            $allProcesses = @(Get-CimInstance Win32_Process)
```
UltralyticsAssistant, line 161: 💡 **MEDIUM**: The Windows cleanup still relies on a single parent-PID snapshot rather than OS-owned containment. If an intermediate child exits after creating a long-lived grandchild, that intermediate PID is absent from `$allProcesses`, so the traversal cannot discover the surviving grandchild; if the root has also exited, `taskkill /T` cannot reach it. The wrapper then returns 124 and starts the retry while the process remains alive. Use a Windows Job Object or another supervisor-owned containment mechanism that remains valid as descendants detach or intermediate processes exit.

retry/action.yml
```diff
@@ -126,22 +131,135 @@ runs:
...
+                for pid in children.get(pending.pop(), []):
+                    if pid not in found:
+                        found.add(pid)
+                        pending.append(pid)
+            return found
+
+        def tracked_processes(token):
+            """Return processes carrying the attempt's unique inherited environment marker."""
+            marker = f"ULTRALYTICS_RETRY_TOKEN={token}"
+            processes = subprocess.check_output(["ps", "axeww", "-o", "pid=", "-o", "command="], text=True)
```
UltralyticsAssistant, line 209: 💡 **MEDIUM**: On macOS, descendant cleanup still depends entirely on the inherited `ULTRALYTICS_RETRY_TOKEN`. A command can detach and scrub its environment (for example, by launching a descendant through `env -i`); that process then disappears from `tracked_processes()`, while the macOS path does not perform the Linux subreaper/descendant scan. The wrapper kills only the original process group, returns 124, and starts the retry with the escaped process still running. The README documents this limitation but does not prevent the process-tree leak; use OS-owned containment or another tracking mechanism that the child cannot remove.

UltralyticsAssistant (commented):
## 🔍 PR Review 7

Since review 6, the Python prerequisite is now explicitly validated and documented, and Windows cleanup performs descendant discovery before killing the root. The per-attempt timeout and status handling look correct, but Windows parent-PID cleanup and macOS environment-marker tracking still allow detached descendants to survive into retries. Not LGTM.

- 💡 **MEDIUM** `retry/action.yml:161` The Windows cleanup still relies on a single parent-PID snapshot rather than OS-owned containment. If an intermediate child exits after creating a long-lived grandchild, that intermediate PID is absent from `$allProcesses`, so the traversal cannot discover the surviving grandchild; if the root has also exited, `taskkill /T` cannot reach it. The wrapper then returns 124 and starts the retry while the process remains alive. Use a Windows Job Object or another supervisor-owned containment mechanis…
- 💡 **MEDIUM** `retry/action.yml:205` On macOS, descendant cleanup still depends entirely on the inherited `ULTRALYTICS_RETRY_TOKEN`. A command can detach and scrub its environment (for example, by launching a descendant through `env -i`); that process then disappears from `tracked_processes()`, while the macOS path does not perform the Linux subreaper/descendant scan. The wrapper kills only the original process group, returns 124, and starts the retry with the escaped process still running. The README documents this limitation but…
</details>

UltralyticsAssistant:
🎉 PR [#879](https://github.com/ultralytics/actions/pull/879) has been merged—thank you, @glenn-jocher!

This update brings those improvements together with per-attempt timeouts, process-tree termination, and expanded cross-platform tests, allowing configured retries to proceed when commands hang.

